PR #135, opened 2015-05-23
Rotworm OSI Style

ppardalj:
-Adds blood disease ability to Rotworm
-Adds special effect for Candlewood Torch
-Adds Writs of Membership drop to Rotworm


prostagma:
Hi there, 
I saw that in UO Guide the value of the Str of a Rotworm is 222 - 277. The value of the Str of the Rotworm in the code was set to a random value from 200 to 300. Does it matter?

ppardalj:
Info from UOGuide is not always accurate, it is based in few samples on OSI but sometimes values may be wider. I've seen rotworms with actual values of STR lower and higher than the ones listed in UOGuide :)

prostagma:
Nice!

PR #207, opened 2016-01-02
Cleaned up Gargoyle Citidel decorations and signs

### LINKED ISSUE #202: Old Gargoyle Citadel Signs
The signs from the old Gargoyle Citadel need to be moved to old.

qbradq:
Fixes #202


dmurphy22:
Why was the directory changed?

qbradq:
I did it this way to preserve the decoration files for folks that want to run an older era, like how we still have the Ruined Magincia and Old Haven files.

WarriorWithin:
also Ilshenar Blackthorn castle should be cleant from wandering doors and so on

PR #312, opened 2016-01-11
Ninja AI

WarriorWithin:
I created a dynamic Ninja AI, it's close to pvp, maybe a bit too much?


qbradq:
Thank you for the contribution, but we try to keep things OSI-like in ServUO. I understand that our Ninja AI doesn't match that of OSI and it is something that needs to be addressed, but this does not appear to move in the correct direction. I think it will be great fun for your shard though!

Issue #329, opened 2016-01-15
Medusa Susceptible to Dryad Allure Spell


qbradq:
Should still be susceptible to Repond Slayer, but not this spell

Ravenwolfe:
Hey, I addressed this in JustUO by adding a AllureImmune property to basecreature. I can send you a pull request for that if you want.

qbradq:
Resolved by pull request #330. Thank you Ravenwolfe!

PR #330, opened 2016-01-15
Issue #329 Adds allure immunity to certain mobiles and the ability to make any m…

Ravenwolfe:
…obile immune.


qbradq:
Noting that Dryad's Allure is currently disabled. Anyone know why? I am going to have to do some more detailed QA before I turn it on.

qbradq:
Tested Dryad's Allure on several mobs both humanoid and non and it worked as expected. Looked through code and stabling and hitching checks are there, although I did not explicitly test them.

PR #331, opened 2016-01-15
Non-OSI system commented out

Ravenwolfe:
XMLPoints and Mob Factions are not default OSI behavior and should be disabled by default.


dmurphy22:
Agreed.

PR #333, opened 2016-01-17
A better system to turn on and off XMLSpawner extra systems and also …

Ravenwolfe:
…disabled them by default since they are non-osi.


Ravenwolfe:
Actually, after I sent this, I realized that you already had a shard configuration script that pulls from cfg files. You are welcome to use what I did or just incorporate it into the shard configuration if you think that would be better!

qbradq:
I liked this pr. I was just going to swap out the xml based config stuff. Are you doing a new version, or shall iI?

Ravenwolfe:
If you got time, go for it, I just got a major project at work.

PR #383, opened 2016-02-06
Fix for Scripts/Mobiles/Personal Attendants/Guide.cs not finding

anasiati:
Fix for Scripts/Mobiles/Personal Attendants/Guide.cs not finding Data/Guide/Definitions.cfg (Linux only?). Path was windows dependent.

### DIFF
--- a/Scripts/Mobiles/Personal Attendants/Guide.cs
+++ b/Scripts/Mobiles/Personal Attendants/Guide.cs
@@ -58,7 +58,7 @@ namespace Server.Mobiles
 
             try
             {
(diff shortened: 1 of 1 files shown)


qbradq:
Thank you so much for finding this issue! Our Linux support definitely needs some polish. Check out System.IO.Path.Combine(). We'd rather use this method for constructing all path names. The way you went about it is basically the same thing, but with more code. The method also works for file systems that have multi-character path separators like zfs in archival mode, although I doubt that is going to come up anytime soon :)

boredcoding:
Awesome, After grepping around I could see Path.Combine is the norm. I'll make sure and squash these up in advance next time.

PR #428, opened 2016-02-18
Add - Config file for PlayerCaps and the needed file changes in the …

pyro17:
Add - Config file for PlayerCaps and the needed file changes in the whole project 
Refactoring - veteranreward config
Fix  - Older expansions still didnt have their default full bonus skillcap applied added if to solve this


qbradq:
All of these commits have been merged in #447

PR #479, opened 2016-02-26
Random Item Generator and Despise Updates...

kevin-10:
-Took out redundant Random Item Generation in Loot.Construct
-Added GetLuckChance() method to Lootpack in the event you want a value for luck from the luck itself, not just the mobile
-Uncommented MageArmor and added Elemental Bonuses via AosWEaponAttributes to Imbuing Table.  This is the table that is used when calculating item intensity in Imbuing as well as Runic Reforging/Random Item Generator. This will prevent too many mods on weapons
-Various fixes/tweaks to RunicReforging.cs:
-Fixed type cast that was allowing resists to exceed the cap of 15
-Fixed a couple min/max intensity calculation to be more EA like
-Fixed issue where no properties were being applied to items
-Reactive paralyze now only appears on prefix/suffix items that are either BaseShield or two handed weapon.


qbradq:
PR was merged with the the mentioned changes to feature compatibility. It was test by generating 100 daggers, leather caps, floppy hats, and golden bracelets. Nothing appeared out of the ordinary. I am keeping the system disabled by default until we have time to vet it on the ServUO production shard.

kevin-10:
Hey norm,

I see your recent commits for loot.  I still think it's a bad idea to force items  through the random item generator through Loot.construct(...). If you want a specific item to pass through the random item generator, it should be done on those scripts. For example, monster loot should be called in Lootpack.cs. If you want it to drop in treasure chests, then You can put a hook in treasurechest and/or treasuremapchest as well as the appropriate expansion checks. If it doesn't pass the expansion check, then it's handled by the old runic tool way.  

Loot.construct is a convenient way to construct items from a type, and forcing item properties on them can become problematic in the future of you don't want those properties.

PR #492, opened 2016-02-29
Fix for Gargoyles taking damage while flying

SirGrizzlyBear:
Fix for Gargoyles taking damage while flying
#no issue number


qbradq:
Thank you for the PR! Is this how flying works on OSI?

SirGrizzlyBear:
This is per OSI.  Tested by flying around on OSI Sonoma vs, play.servruo  vs my test server.  No damage is taken on OSI where damage is taken on play.servuo.

PR #496, opened 2016-02-29
Imbuing Pass

SirGrizzlyBear:
First Pass at getting imbuing to be more inline with OSI.  Cleaned up the Unravel code and imbue last item.


SirGrizzlyBear:
This also fixes issue #507

Issue #459, opened 2016-02-23
Paragons in Ilsh will attack players when they should not due to karma. 


dmurphy22:

qbradq:
Fixed in #459

PR #509, opened 2016-03-06
Fix for #459

SirGrizzlyBear:
Fixes Paragons in Ilsh will attack players when they should not due to karma.


qbradq:
Let me know your forum name for the contributors list.

SirGrizzlyBear:
SirGrizzlyBear on Forums

PR #569, opened 2016-03-16
Special Fishing Net and Kraken

SirGrizzlyBear:
Updated the loot rate / items for wild spawn krakens to mach OSI.  Tweaked the spawn and rates on Special Fishing Nets to mach OSI.  Burned through a stock pile of Special Nets on Sonoma testing. and ref. http://www.uoguide.com/Kraken


qbradq:
Thanks so much for this!

PR #574, opened 2016-03-18
Attempt at making Doom artifact gain like Pub48

SirGrizzlyBear:
Ref. http://www.uoguide.com/Publish_48

Setup doom artifact drop system like ToT.  Point generation cut in half, ideal guaranteed drop is 9 to 10 clears.  Testing i averaged 5 to 7 clears till a drop.

Todo: Figure out how to implement the top damager getting bonus points.


Scripts/Mobiles/BaseCreature.cs
```diff
@@ -5260,6 +5260,10 @@ public override void OnDeath(Container c)
 							givenToTKill = true;
 							TreasuresOfTokuno.HandleKill(this, ds.m_Mobile);
 						}
+                        if (Map == Map.Malas || region.IsPartOf("GauntletRegion"))
```
qbradq: This would be &&, not ||, correct? Otherwise you'll be calling this logic for every kill in Mala, which isn't necessary.

qbradq:
Great PR! Thanks so much! Just one thing I noted in the code review. Please have a look at BaseCreature.

qbradq:
Thank you so much for the contribution!

PR #637, opened 2016-03-28
Fix for animaltrainers pets and couple missing shoes

GenerationOfWorlds:
Fixed Animal trainer pets to not follow when you buy them per UO
Accurate
Added couple missing shoes as well.

### DIFF
--- a/Scripts/Mobiles/Vendors/BaseVendor.cs
+++ b/Scripts/Mobiles/Vendors/BaseVendor.cs
@@ -27,7 +27,9 @@ namespace Server.Mobiles
 		Shoes,
 		Boots,
 		Sandals,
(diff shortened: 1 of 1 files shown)


qbradq:
Having animals follow when bought from an NPC is a convenience for our players. Folks that don't normally play tamers often forget to have their horses follow them, and figure that out when they dye and and the horse won't follow their ghost.

Also, the shoe commit won't work. You have to edit the random generation code as well to have them populate themselves. As-is the vendor will be bare-foot if the new values are selected. Also we wouldn't want the SE foot ware outside of Tokuno.

I like where it's going though!

dmurphy22:
Request denied. Reason: User is banned from the community.

PR #638, opened 2016-03-28
PS and Chest Changes

Muddled:
Changed lockpicking skill values to OSI values. They were previously unpickable due to being higher than 100. Added SA skills onto the powerscroll list.


qbradq:
Thanks so much friend!

PR #654, opened 2016-03-28
AI Changes and Weight Change

Muddled:
Removed Revenant from NecroMage AI and Mage AI due to it not working correctly in trammel facets. Fixed Daemon Bone weight as it's now considered a reagent.


qbradq:
Why are you removing these necro spells from the AI rotation?

qbradq:
Also need to update your branch with the taxonomy changes to resolve the merge conflict.

PR #663, opened 2016-03-29
Adjusting Ancient MIB Chance

SirGrizzlyBear:
Better formula for a random 1 in 25 chance.  was not sure if the other
method would actually ever be true


qbradq:
I agree that the previous code was a little non-obvious. Usually we say Utility.Random(25) < 1. Utility.Random(25) == 0 also works, so does this. Whatever floats your boat :)

PR #665, opened 2016-03-29
Updating SoS chests to work closer to OSI.

SirGrizzlyBear:
Players should be able to have a chance to get a Arty, and a decor /
tmap.  This chance would be super rare, but it should be possible.  Also
moved the creation of Deco items outside of the bag as they appeared in
the base chest along with the maps that should have a possibility of
appearing.  Also removed deco items from arty items as they should be
separated and not take away from getting an arty.


qbradq:
Can you update your branch with master? We've moved TreasureMapChest.cs today and that's causing the conflict.

SirGrizzlyBear:
updated woo no conflicts!

PR #670, opened 2016-03-30
Attempt to fix Hanks tamer bug

SirGrizzlyBear:
Ran through Champ spawn with greater dragon.  Tried all methods of getting pet to go grey based on reports.  Could not reproduce with changes made.


qbradq:
Interesting :) Thanks! I think others were having trouble in trammel ruleset maps with thier pets just refusing to attack or guard due to this.

PR #718, opened 2016-04-07
Fix for mage armor appearing on clothing

kevin-10:
- convenient methods for getting aos attributes from an item
- Self Repair will no longer appear on brittle, antique and no repair items
- Fixed Imbuing property not showing on pre-patch items
- Tram-Fel maps now spawn in random locations


qbradq:
This PR is OK except for the Trammel / Felucca TMap wrapping. It needs to be setup to avoid areas near towns and dungeon entrances, especially island towns like Moonglow and Skara Brae. The first incarnation of the TMap system had a few map locations on the main islands of Moonglow and Skara Brae, which eventually resulted in Balrons roaming the islands and whacking unsuspecting players. Those locations were removed which is why we ended up with less than an even 200 locations.

Can you update the wrapping so that it only selects appropriate areas?

kevin-10:
Will do.  Ill look at getting it in today after work.  So, for the wraps, I will only exclude the inhabited islands.  Towns already have restrictions on the Point2D generator.  As far as whacking unsuspected players, that's going to happen regardless.

dmurphy22:
I am gonna pull this PR and we can work on the map locations at a later time.

PR #719, opened 2016-04-07
Fixes an exploit used to access trammel.

jackuoll:
Steps to reproduce this:

- Go to Felucca, stand in an inn (or any spot you can insta-log out)
- Use the stuck menu and select an option to get the stuck timer running.
- Log out, wait 2 minutes for the stuck timer to reach its OnTick method.
- Log back in, you're standing in trammel.

This would also cause logged out players to be permanently stuck in the world until they next logged in or the shard restarted.

Revert was done because I was going to remove redundant using statement, but it included the revert in the commit log. Meh.

### DIFF
--- a/Scripts/Services/Help/StuckMenu.cs
+++ b/Scripts/Services/Help/StuckMenu.cs
@@ -1,6 +1,8 @@
 using System;
 using Server.Gumps;
 using Server.Network;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


qbradq:
Thank you for the patch! You're a rock star for finding this :) I'll get it merged as soon as I get it tested on my end.

PR #766, opened 2016-04-23
Master

mithril52:
Some fixes of Race restriction removal for now


mithril52:
Sorry! Went to the wring place by accident!

PR #833, opened 2016-05-21
Abuse Fix

Argalep:


ShadwDrgn:
This is not actually abuse on ToT. Official shards allow you to farm champ spawn mobs for ToT arties. This can be confirmed by googling about how to farm tot artifacts. Almost every guide suggests the sleeping dragon champ spawn

PR #950, opened 2016-08-20
More bug fixes!

kevin-10:
- Fixed FC issue with Protection spell active
- Fixed messages for new fish when they are fished up
- Fixed fishing gains


kevin-10:
THis one will require a core re-compile.

PR #1000, opened 2016-08-30
- Fixed TItles menu issue with overhead Title

kevin-10:
- Mantle of the fallen now has cast focus
- Commodity deed box is now flippable via decorator tool
- Chest of sending now has 50 charges to Start
- Fixed crash with disguise kit
- Fixed/refixed issue with pets moving wayyyyyyy tooooooo sloowwwwwwwwwww when dead
- Fixed honesty item propery being removed too early (needs core recompile)


kevin-10:
I'm not sure what I did wrong here.

PR #1100, opened 2016-10-14
Blackthorn Point Fix

Argalep:


kevin-10:
The original hitching post is a craftable.  You will have to create, derive, or modify as to where the original hitching post will still work in houses.  I would suggest creating a class for uses outside of housing.

PR #1103, opened 2016-10-15
Blackthorn etc

Argalep:


kevin-10:
Whats the deal with the damage/relocation/poisoning?  I don't see that in uo.com.

Also, the code for dismounting is pretty incomplete and has an invalid cast which will cause crashes.

Argalep:
Characters who hide in the dungeon will be auto-revealed after a short interval. Characters who remain in one place for too long will be auto-teleported to another location in the dungeon. This location is usually in the middle of a poisonous gas cloud surrounded by hostile creatures which will immediately attack the character. 

http://www.uoguide.com/Castle_Blackthorn_Dungeon

I checked OSI test server

kevin-10:
I put in the Agent of the Crown changes that you put up.

PR #1119, opened 2016-10-20
Generate Exodus

Argalep:


Tasanar:
I know on real UO they also clean up untamed pets. However it is done either after 12 hours or after a shard restart (they do theirs once a day). I say this because I have had a pet go untamed logged in a few hours later and it is still there.

I only say this because some of us save our shards hourly. I would just suggest you use a Utc.Now and maybe make it once every 12 hours and not after every save. That is a little harsh.

Otherwise awesome addition!

PR #1141, opened 2016-10-31
Update BaseAI.cs

GriffonSpade:
if (bFacFriend && !m_Mobile.IsFriend(m))
Should have been a nested check. New change renders it moot.

New Change makes bFacFriend and bFacFoe fully binary instead of previously (intended) bFacFriend overriding bFacFoe when the former is true.


kevin-10:
Nice work.

PR #1159, opened 2016-11-09
*add Mini Champion System & Labyrinth Spawns

Argalep:


Tasanar:
Awesome! I hate XMLSmart spawning and when I global change times or turn it on and off I sometimes mess these mini champs up. Currently they run off of XML now they will have a set script.

PR #1211, opened 2016-12-04
Unravelling Imbuing Gains

Vorspire:
+ Potential fix for Unravelling skill gains and internal resource item leak.

### DIFF
--- a/Scripts/Skills/Imbuing/Core/Imbuing.cs
+++ b/Scripts/Skills/Imbuing/Core/Imbuing.cs
@@ -601,101 +601,108 @@ namespace Server.SkillHandlers
             }
         }
 
-        public static bool UnravelItem(Mobile from, Item item, bool message = true)
-        {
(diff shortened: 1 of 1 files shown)


Ravenwolfe:
Good fix for internal items but caused two new issues:

1. Its giving you resources based on your skill and not the items itself. When I unraveled an item at 50 skill I got magical residue. When I unraveled it at 80 skill, I got essense. It should still give me residue based on the item, otherwise you would never be able to get residue at higher levels.

2. Its not correctly counting the number of items in a container and therefore not unravelling them all at once. Example, put an item in a bag, dupeinbag 100, then unravel. It does like 20 items and leave 80, forcing you to do it multiple times.

I tested the old way and the unravelling and skill gain worked fine, I never should have posted that issue as the report to me was wrong. Perhaps leave it the old way but incorporate the internal item fix and cleanup?

PR #1220, opened 2016-12-07
Add and Fix Berserk Function

Argalep:


kevin-10:
Can we create a BeastialSetHelper to handle this functionality, instead of modding up PlayerMobile.cs for one item set?

kevin-10:
Is there a reason why Mobile HueMod can't be used?  Also, instead of all that code in PlayerMobile, how about adding a simple hook, and handling it in the suit helper?

Argalep:
Body color changes when full armor is worn

PR #1236, opened 2016-12-14
*add Berserk Mode

Argalep:


kevin-10:
Moving the functionality from PlayerMobile.cs to Mobile.cs (core edit) isn't what I had in mind.  The set helper can handle all of this. 

- Static methods to evaluate how many set pieces, appropriate hue, etc
- Static method hooks in AOS.cs for damage modifiers, etc (i think you have this already)
- Table (Dictionary) to handle all of the instances of timers when suit goes active

Argalep:
Affects all set pieces worn over the character.  Color Range (1255 - 1260)

PR #1278, opened 2016-12-20
Berserk Hot Fix

Argalep:


kevin-10:
The bestial suit beserk and race bonus beserk are completely different.  The suit changes hue of the suit and player wearing it, and absorbs damage based on the beserk level and total suit pieces being worn.  The race ability increases weapon/spell damage output and does not hue anything.  The suit can easily be handled in  the BestialSuitHelper without core edits to Mobile.cs, etc.  The only distro edit you need is in AOS.cs.

kevin-10:
Also, just FYI, Gargoyle Race ability beserk is already properly functioning.

PR #1336, opened 2017-01-02
*add Ortlem the Mystic Quest

Argalep:


ppardalj:
Please add credits: https://github.com/xrunuo/xrunuo/blob/master/Scripts/Distro/Items/Quests/Misc/CrystalBallOfKnowledge.cs

Argalep:
I added it as a reference.

PR #1496, opened 2017-01-22
Update Navrey.cs

Argalep:


kevin-10:
a null check will need in case navery is every spawned standalone.

PR #1521, opened 2017-01-26
Bug Fixes

kevin-10:
- Added support for morph earrings
- Fixed slayer drop as regular loot, no more eodon slayer
- Added 100 tile range check to champion for powerscroll drop, including protector
- decreased drop chances for SOT on Navery
- Fixed GrizzlyBear turning into RagingGrizzlyBear
- Pirate/Merchant ships now must be scuttled to board them
- Fixed issue where ghosts were given credit at the VvV Altar
- Added Config settings for new loot generator
- Fixed peerless master key issue where it would not let players back in the boss encounter


maicomstoppa:
Wow. That was fast! Awesome work! =D

PR #1565, opened 2017-01-29
No unsafe

maicomstoppa:
This is causing some MONO servers to stuck at saving world. I coudn't measure any impact in performance.


Vorspire:
This will need to be reviewed for consistency. 
There may be inconsistencies in byte endianess that need to be compensated for.

maicomstoppa:
Sure, no rush.

I have mixed feelings about repoing this too. As I see, the only people benefiting from this are the ones trying to run ServUO on the MONO ARM version: Raspberry PI, C.H.I.P, etc, which requires aligned floats.

More info: http://www.aleph1.co.uk/chapter-10-arm-structured-alignment-faq

I do think that cross-platform is a very important aspect of ServUO, but not at the expense of performance/security/wathever for the people that are already using this on shards over the planet.

But, if stated that this brings no harm to the current server quality, repoing this would be awesome, since people could make more servers using inexpensive hardwares like the C.H.I.P I'm using ( U$9 ), Orange PI ( starting at U$ 7 )  and at (virtually) no energy costs.

pyro17:
You could also make a compile condition so if it has the ARM tag it would use your version else the old one.  same way how mono specific parts are handled

Vorspire:
Going to close this PR on the grounds that it has a niche target audience and may cause unforeseen issues for those outside of that scope. That is not to say these updates cannot be implemented eventually, but I'd like a degree of certainty from the author before doing so.

PR #1582, opened 2017-01-30
Update zLib from 1.2.3.0 (2005) to 1.2.11.0 (2017)

Xeroxxx:
Update zLib Library

License allows redistribution without acknowledgement.
http://www.zlib.net/zlib_license.html

>  1. The origin of this software must not be misrepresented; you must not
     claim that you wrote the original software. If you use this software
     in a product, an acknowledgment in the product documentation would be
     appreciated but is not required.
 > 2. Altered source versions must be plainly marked as such, and must not be
     misrepresented as being the original software.
>  3. This notice may not be removed or altered from any source distribution.

Additional notes (from readme.txt of zLib)
----------------
**- The new DLL was renamed because there exist several incompatible
versions of zlib.dll on the Internet.**

- This DLL, named zlibwapi.dll, is compatible to the old zlib.dll built
by Gilles Vollant from the zlib 1.1.x sources, and distributed at
http://www.winimage.com/zLibDll
It uses the WINAPI calling convention for the exported functions, and
includes the minizip functionality. If your application needs that
particular build of zlib.dll, you can rename zlibwapi.dll to zlib.dll.


Xeroxxx:
zlibwapi32.dll: https://virustotal.com/de/file/5daaaf040f038f5b77d6ed33c031f3851604cb1dab89b074ff92893b3a60d3fc/analysis/1485793359/

zlibwapi64.dll:
https://virustotal.com/de/file/fcf9f36b66fd053e4bfe4c599d5bd37bc66d6891a7af8616079786b942f316f3/analysis/1485793416/

PR #1609, opened 2017-02-02
EC Client Support & Fixs

Argalep:


maicomstoppa:
THIS. IS. AWESOME.
=D

Hey, argalep!
Just to expose my findings, after the hotbar, face and grid containers fixes, the only things I saw missing in EC were:

1) Can't rename runes, use the tools "Make Number", or say "unorus" to initiate the Sacred Quest as ServUO can't read from ec input: should be a gump.

2) For some HUEd mobs we have to put on their spawner the "/hue/xxx" thing, because the EC uses the same graphic (this is a supposition) than the regular one. Example: The Fire Steed shows in EC as a regular steed.

Argalep:
Hi maicom,

There is a solution related to menus in the project justuo. Need to integrate

https://github.com/zerodowned/JustUO?files=1

maicomstoppa:
What menus, exactly? 
Do you mean the Context Menus (missing strings)? If you does, they were already fixed in my commit (afaik).

Or do you mean the gumps?

kevin-10:
I had to revert this.  Opening containers, the items were not displaying.  This was happening on EC and Classic Client.  It seemed like a packet issue, as the container displayed the item count and weight properly.

PR #1616, opened 2017-02-03
- EC Support

Argalep:


Vorspire:
Everything looks good, but I have one request before this can be merged...

Some of the core files like Item.cs that have modifications to OnDragDropInto, etc - it would be in the best interests of compatibility to provide overloads for these methods that call the new methods with a default value for the new argument. For example, gridloc can default to 0 or -1. This will ensure that administrators won't need to update all the scripts that override the old methods, as those old signatures will still be present.

kevin-10:
Also, was the issue with the items within a container solved?  If so, where was the issue?

Frazurbluu:
Something else to consider, the new contents of VirtualHair could be in a new file VirtualFace ect..
probably several places where this could be done.. and I only mention it because I was told the same thing when I first came here. :)

Frazurbluu:
Also there were a few fixes made after SE/GOW that may need added..

maicomstoppa:
Guys, can I ask something?

When I was studying about the mods I had to do about the hotbar, I saw several methods to detect the Enhanced/KR Client, and none of them worked for me. So I made the one that is repoed now, "IsEnhancedClient", and it is working.

Could someone with more knowlege than me take a look at it?
I'm asking this because every other method I saw relied on packets that never came for me, and my solution was so simple, so simple, that I'm asking myself if it covers every case, even if it is working (for me).

Thanks!

Vorspire:
This method seems to have worked fine for me in the past, but I rarely deal with EC (I hate it lol)

https://github.com/Vita-Nex/Core/blob/master/Extensions/Server/NetStateExt.cs

kevin-10:
That method should be fine for now, as long as the classic client major never bumps higher than the EC!  For Vorspires method, it would have to be tested.   I have no clue if the EC passes through the stuff to be detected as the 3d client.

Argalep:
hi dexter, 

> Also, was the issue with the items within a container solved? If so, where was the issue?

Container.cs, Item.cs >> Changes to GridLocation

Need to be tested

kevin-10:
Cool, but see the comment about the default arguments.

Frazurbluu:
Looked more over, I guess its from w/e DS took some fixes from thus the similarities or its just updated, some things I hadn't seen added before.. nice! Looking forward to seeing this get revised and added.

Frazurbluu:
@Argelep, any chance you can inform me where this code is from instead of me speculating?
Its my opinion..,  If you didn't write it, you should include its origins in the PR.

kevin-10:
this virutal method:
public override bool OnDragDropInto(Mobile from, Item item, Point3D p)

adds gridloc argument:
public override bool OnDragDropInto(Mobile from, Item item, Point3D p, byte gridloc)

where gridloc should be a default gridloc = 0 or -1.

kevin-10:
Just tested, the default gridloc is 0, so the two methods can be made as follows:

public virtual bool OnDragDropInto(Mobile from, Item item, Point3D p, byte gridloc = 0)
public virtual bool DropToItem(Mobile from, Item target, Point3D p,  byte gridloc = 0)

Vorspire:
I didn't mean to use default parameters, I mean, add an actual overload for those methods that doesn't have the gridloc argument.

Using default parameters is still changing the method signature and that will still cause issues with compatibility.

public virtual bool OnDragDropInto(Mobile from, Item item, Point3D p)
{
    return OnDragDropInto(from, item, p, 0);
}

public virtual bool OnDragDropInto(Mobile from, Item item, Point3D p, byte gridloc)
{
    // stock code
}

Vorspire:
Actually, I take back my last example, because calling the new method would end up not calling the old method at all, and therefore not calling custom user's override.

I know it's a headache to do this properly, which is why most implementations just create an Item.GridLocation property which is update appropriately and independently from these methods like OnDragDrop.

Anyway, I'd recommend a pattern more like this to ensure full backward compatibility:

public virtual bool OnDragDropInto(Mobile from, Item item, Point3D p)
{
// stock code
}

public virtual bool OnDragDropInto(Mobile from, Item item, Point3D p, byte gridloc)
{
// do something special with gridloc

return OnDragDropInto(from, item, p); // call the original stock method.
}

kevin-10:
Or, why not set GridLocation before calling these functions?

PR #1643, opened 2017-02-05
Mastering the Soulforge Reward Item Fix.

Argalep:
http://www.uoguide.com/Mastering_the_Soulforge

*only A Legendary Scroll of Imbuing (120 Skill)


kevin-10:
http://www.uoguide.com/Runic_Tools

PR #1644, opened 2017-02-05
Secrets of the Soulforge Reward Item Fix.

Argalep:
http://www.uoguide.com/Secrets_of_the_Soulforge

* only  A Wondrous or An Exalted Scroll of Imbuing (105 or 110 Skill)


kevin-10:
http://www.uoguide.com/Runic_Tools

PR #1675, opened 2017-02-07
*add [GenWinchAssembly

Argalep:
* ladder location fix


Argalep:
You can use it when you have problems

PR #1684, opened 2017-02-08
2D client vendor context fix

Argalep:


kevin-10:
Will this have any impact on 2D client?

Argalep:
6b336a310f37ee25177194a91e9d4fa03acc3d1144f875efcf pimgpsh_thumbnail_win_distr

kevin-10:
So this is what it looks like now?

PR #1708, opened 2017-02-11
Update Packets.cs

Argalep:
neterr.log fix


kevin-10:
Is that something that is not needed?  Can someone confirm that it's ok to chop that code out?

Argalep:
I added this code myself

https://github.com/ServUO/ServUO/commit/d42a361ef474188e46547d28fe67eecb347bac32#diff-d8331dd843eb30498e1b1ca0dff4a226

Xeroxxx:
Is the type casting to (int) required? Original code was without it.

But yes this part of code is doubled.

PR #1711, opened 2017-02-11
Garbage Collector GCServer or manually Set

Xeroxxx:
GC if GCServer set true OR GConSave is set.

Readworthy: https://msdn.microsoft.com/de-de/library/ee787088(v=vs.110).aspx#workstation_and_server_garbage_collection

capture
capture1
untitled

### DIFF
--- a/Config/AutoSave.cfg
+++ b/Config/AutoSave.cfg
@@ -11,3 +11,8 @@ Frequency=00:00:05:00
 # Default: 15 seconds
 # Format: dd:hh:mm:ss
 WarningTime=00:00:00:15
+
+# Enable Gargabe Collection after WorldSave
+# This maybe useful on large shards
+# Read: https://msdn.microsoft.com/de-de/library/ee787088(v=vs.110).aspx#workstation_and_server_garbage_collection
+GConSave=false
--- a/Scripts/Misc/AutoSave.cs
+++ b/Scripts/Misc/AutoSave.cs
@@ -70,6 +70,7 @@ namespace Server.Misc
 		GC.Collect();
 		if (Core.Debug)
 			Console.WriteLine("Total Memory before Garbage Collection: {0}", GC.GetTotalMemory(true));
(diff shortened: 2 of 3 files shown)


Vorspire:
Doing this on world save can dramatically increase the save times.
When the Pandora shard was using 10gb of RAM (holding over 6 million world items) in a production setting, we attempted to force GC on world save and it almost doubled the overall save times. 
This was the reason I added the [GC command to Vita-Nex: Core, so the owner of the shard could run it manually when he felt the server was lagging.

There are not many shards that have been running for 6+ years like Pandora, but this could pose an issue for shards that have a large world database in the future.

Thoughts?

Xeroxxx:
GC isn't happening ON WorldSave...its happening AFTER it. 

However this is not enabled by default and noone should enable it without knowing what he is doing. Notice is added to config.

I think its quite important to know if pandora was running with GCServer enabled or not, as the behavior is different.

dmurphy22:
This won't be implemented. 

Why it's not being implemented is a very complicated thing. The GC on windows runs when it does for a reason. It's not a random thing and what it does when it does run is not random either. The GC in .net is basically perfect. It knows how much ram it needs to free and when it needs to free it. Calling the GC manually like this is not a good idea. Freeing ram for the sake of freeing ram is not a good thing. We are basically creating a very expensive operation in order to have more expensive operations later. The GC freezes the thread after save, then later ram will need to be allocated to the process again because it has been freed prematurely which is another expensive process and could happen hundreds or thousands of times and will repeat after every save.

TL:DR a few hundred MB of ram usage now is not worth saving the CPU time later, especially since that ram will fill up again anyway.

Xeroxxx:
Mhh okay...actually memory isnt a  thing anymore...we have plenty ^^
However I think it makes sense when running gcserver as the behavior of stopping threads is different, I may add this as a seperate script for my own.
thx for looking into this too

PR #1715, opened 2017-02-12
* new doom arties fix

Argalep:


Vorspire:
Do all of these changes have appropriate era distinction checks?

The changes to m_DoomArtifacts should be accurate to all era's.

kevin-10:
Which would we use?  TOL I would say, even though this change was post-TOL.

PR #1741, opened 2017-02-18
I tried to make Ultima.Dll not static

Xen85:
Unstaticize methods in Ultima.DLL because it could be used differently in different programs and because in this way it's possibile to allow more the one client folder at time.


Xeroxxx:
Lots of unnecessary files.

Does not compile.

Xeroxxx:
25 Warning
4 Errors

dmurphy22:
I'm not sure of any practical uses of this? Can you tell me what you think this could be used for?

Vorspire:
Ultima.dll supports multiple data paths to client files.

It is not unreasonable to manipulate the static list that contains the data paths, before calling the required methods.

Static code is also much faster than non-static code, which is essential when processing the game files and their contents.

The Ultima SDK can be wrapped in a non-static context if instanced usage is required, otherwise, there is no need to do it in the context of ServUO. Doin so would also increase memory usage, as the caches would need to be duplicated on a per-path basis.

PR #1742, opened 2017-02-18
Introducing Cache

Xeroxxx:
Introducing the Cache.

When the Server is started the first time:
if art files are available a cache folder is created and generating Bounds.bin
if not Data/Binary/Bounds.bin is used
if Cache/objects.xml is not available its been generated
deleting cache folder forces regeneration
deleting cache.cs completely disables the cache and default files are loaded.

Why this makes sense?

Art files are updated by osi and custom shards regulary, so the new bounds can be generated manually and it does not require a ServUO Developer to regenerate the file regulary.

Alternative would be:
Adding Rebuild Bounds.bin to AdminGump Maintenance Category and rebuild it in place.

https://github.com/ServUO/ServUO/pull/1745

Simply reject the one you didn't like.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 /Scripts/Output
 /Logs
 /Backups
+/Cache
 /Saves
 /Web
 /Docs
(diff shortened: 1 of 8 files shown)


dmurphy22:
Superseded by #1745

PR #1749, opened 2017-02-19
Reorganization of Filestructure. Moving ChampionSpawns.xml to appropr…

Xeroxxx:
…iate folder "Spawns"


dmurphy22:
Spawns folder is for xmlspawner files only.

PR #1762, opened 2017-02-20
Exploring The Deep Quest Mobile dropQuest Item Fix

Argalep:


kevin-10:
Should the item creation be inside the loop?  If not, you're bouncing the same item around to each mobile in the foreach loop.

PR #1767, opened 2017-02-21
Outputs (Debug) adjustment

Xeroxxx:
StaffBook output only on debug
Reports loading colored


dmurphy22:
The current behavior is just fine IMO.

PR #1769, opened 2017-02-21
Security Warning 

Xeroxxx:
Security Warning if ServUO is being run as Administrator or root.

This is not necessary nor can this be recommended. Message is the same, only looks different on screenshots.

root

administrator


dmurphy22:
It is recommended to run ServUO in elevated permissions mode, at least on windows. There are a few methods that will not run correctly if the server is not in Admin mode on windows and the client files are in the C directory.

Xeroxxx:
Is this the only case when admin permissions are needed? Actually I think this is something the users need to get used to. Running Server Software in admin context is a thing that absolutely can't be recommended. Just in case there is an arbitrary code execution issue with servuo...major disaster!

PR #1800, opened 2017-02-26
Juonar, Virtuebane, Anon and Ozymandias updated by uoguide

bsenyuva:
Hits, resists, damage types and skill values updated according to:

http://stratics.com/attachments/roof-bosses-jpg.39702/
http://www.uoguide.com/Virtuebane
http://www.uoguide.com/Anon
http://www.uoguide.com/Juo%27nar


bsenyuva:
Hitpoints of Ozymandias also changed to 9500-10000

PR #1817, opened 2017-03-02
Fixs.

Argalep:


kevin-10:
Why a HashTable for the mephitis web attack?  You  can just use a list.

Argalep:
I corrected

kevin-10:
Sweet, thanks!!!

PR #1818, opened 2017-03-02
Only gargoyle can convert VvV item

Argalep:


kevin-10:
Have you confirmed you have to be a gargoyle?

Argalep:
Yes i checked it

kevin-10:
good, thanks.

PR #1824, opened 2017-03-03
Bug fix PigmentHue of NaturalDye overridden by PlantHue.

uomamoru:
Problem as PigmentHue of NaturalDye become Plain at first time after server startup.


uomamoru:
Oh, sorry. I was bit confused. I mean this problem is about NaturalDye.

kevin-10:
This was fixed a while back.

PR #1831, opened 2017-03-05
Myrmidex Invasion fix

Argalep:
* Creature corpse delete fix
* Tribe spawns only BritannianInfantry


kevin-10:
I already had a fix for the corpses deleting.

Argalep:
Creature, the player is killed, the corpse is not deleted

kevin-10:
They are deleting.  Also, you can flag the monsters as AlwaysAttackable to true and they will always be gray.  The items are blessed, however is you simply mark them as Movable false in the default arg of SetWearable, they will not show up on the corpse.  There is really no reason to set them as blessed.

PR #1833, opened 2017-03-06
Myrmidex Invasion fix

Argalep:
* BritannianInfantry fix
* Tribe spawns only BritannianInfantry
* flag fix location
*  creature vs creature corpse delete


kevin-10:
So the corpses only disappear if they are killed by another creature?   What if a players pet kills it?

Argalep:
Corpse will not disappear when player or his pet kills. I would appreciaye if you can handle the other issue

kevin-10:
Disregard, notoriety has to be handled the way you did, Region only handles harmful/beneficial actions.

kevin-10:
Looks good, but one issue with IsEnemy.  The allegience check needs to be first so they attack players not aligned.  You have it returning false if its a player before the allegience check.

PR #1839, opened 2017-03-08
Bug Fixes

kevin-10:
- Vendors no longer take gold from locked/trapped containers
- Added delay to delete corpses in invasion region to ensure corpse has been created
- Quest items can no longer be toggled while withing a sub pack
- Wildfire items now delete on server Start

### DIFF
--- a/Scripts/Mobiles/NPCs/BaseVendor.cs
+++ b/Scripts/Mobiles/NPCs/BaseVendor.cs
@@ -1379,46 +1379,9 @@ namespace Server.Mobiles
 				totalCost -= discount;
 			}
 
-			if (!bought && cont != null)
+			if (!bought && cont != null && ConsumeGold(cont, totalCost))
 			{
-				if (totalCost <= Int32.MaxValue)
-				{
-					if (ConsumeGold(cont, (int)totalCost))
-					{
-						bought = true;
-					}
-				}
-				else
-				{
-					var items = cont.FindItemsByType<Gold>();
-					var total = items.Aggregate(0.0, (c, o) => c + o.Amount);
-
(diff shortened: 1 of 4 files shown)


Tasanar:
I have a few custom vendors that sell veteran rewards for gold...a LOT of gold. Before Vorspire made changes it was possible  to get the vendor to go into the negative thus resulting in players not only getting all the items for free but also free gold, by the hundreds of millions. Zerodowned was the one who caught it on my shard.

kevin-10:
Regarding the update, I was using the method FindItemsByType provided in the container base class.  Doesn't this return an array as a separate copy, which would prevent a modified exception?  Also, the consume functionality of my custom method was taken from the consume method in the containers base class, it just simply has some checks for locked/trapped containers.

Vorspire:
Sorry, you are correct about the item iteration!
I am so used to working with Find*() methods returning an IEnumerable vs Get*() methods returning a new collection that I overlooked the fact FindItemsByType returned a new collection.

Thanks for correcting me 😎

kevin-10:
The new method should work exactly like the old functionality, except the checks for a locked/trapped container.  I'm assuming this old issue was corrected?

Vorspire:
It uses a double type to prevent integer overflow, so yes, the issue is fixed.

kevin-10:
So vorspire, are we still using your updates you commuted?

Vorspire:
Yes, just assume that an approved PR needs no more input from the reviewer and the merge responsibility defers to the contributor.

kevin-10:
Quick question, how do I take a PR for testing without committing it to the master branch?

PR #1840, opened 2017-03-08
Battle Mobile out of screen frozen

Argalep:


Argalep:
I checked the osi. Creatures are frozen when they are far away from the creatures. You can check it yourself. If there is a different method to freeze the creatures, I will share it with you.

Vorspire:
Why is the white-spacing so inconsistent with these files? :/

Conflicts will need to be resolved and then it's good to go.

Argalep:
WaveStatus : wave not spawns of frozen creatures

Wave does not start with the last update.

kevin-10:
So, on EA, their AI is sImply being disabled if a player isn't within so many tiles?

Vorspire:
I'm not even sure why it's even necessary to do this freezing of all mobs when players are not around... AI's already automatically deactivate when the sectors around it are empty.

Argalep:
Yeah, you can think of it as stopping the war temporarily.

kevin-10:
I disabled the AI deactivation because I figured that once they spawn, they (the creatures) fight each other.  Keep in mind, once the spawn is cleared, they don't respawn until a player comes into the region.

kevin-10:
Let me look into if we can be as flexible about deactivating AI since that is all that is happening.

Vorspire:
If the purpose is to pause the invasion (via the WaveStatus property), then wouldn't it be possible to simply iterate the invading creatures without freezing them? Their AIObject property can be used to determine if their AI is active or not. This should negate the need for iterating players in the region and checking ranges.

Vorspire:
Having the AI deactivate is a normal feature of the AI when players are not in range, unless their PlayerRangeSensitive property is false, then they continue to be active.
Making sure the specific spawns for the invasion have PlayerRangeSensitive set to true, or using their ApproachWait and ApproachRange properties, may also help to "freeze" them.

kevin-10:
I will look into this.  I think its better to use existing code than re invent the wheel.

kevin-10:
looking into it further, their AI does not deactivate because they use waypoints (or actually navpoints).  Why dont we override them to de-activate so if nobody is in the region, they just stand there.  I don't see why they have to be in viewing distance to re-activate.

PR #1853, opened 2017-03-12
ExploringTheDeep questitem fix.

Argalep:
WoodenKeyBarrel respawns fix

### DIFF
--- a/Scripts/Services/ExploringTheDeep/Items/WoodenKeyBarrel.cs
+++ b/Scripts/Services/ExploringTheDeep/Items/WoodenKeyBarrel.cs
@@ -113,7 +113,7 @@ namespace Server.Items
         public override void Serialize(GenericWriter writer)
         {
             base.Serialize(writer);
-            writer.Write((int)0); //version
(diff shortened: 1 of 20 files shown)


kevin-10:
WoodenBarrelKey needs to properly version the new properties added in deserialize.  Any existing items will be deleted, unless its a temporary quest item, in which case increasing the version is not needed.

Scripts/Services/ExploringTheDeep/Items/WoodenKeyBarrel.cs
```diff
@@ -104,13 +113,19 @@ public override bool OnBeforeDestroyed()
...
+
+            writer.Write((int)this.m_key);
+            writer.Write(this.m_StorageLocker);
         }
 
         public override void Deserialize(GenericReader reader)
         {
             base.Deserialize(reader);
             int version = reader.ReadInt();
+
```
Vorspire: Needs a version check here, if less than 1, do not read the new properties - give them default values instead.

PR #1864, opened 2017-03-16
Fixs & Wrong Rewamp

Argalep:
- StorageLocker Fix 
- SorcerersRewardChest respawn fix.

**Add new wrong dungeon** (GenWrongRewamp) 
(thanks demented and RedBeard)

- BedrollSpawner
- Wrong Mobiles
- Wrong Decorates
- Wrong Teleporters
- Wrong Steables Artifact
- Wrong Lvl 3 puzzle 

/*----------------------------------------------------------------------------------*\

**Missing Functions in Wrong Dungeon** 

- http://www.uoguide.com/Ghost_of_a_Prisoner
- http://www.uoguide.com/For_Great_Justice!
- Missing Enchanted Hot Item feature in weapon and armor in trap chest (Wrong Level 3) (Escape from dungeon to remove spell)

Wrong Level 3 videos 
https://youtu.be/pxBohU5vcHM

(Escape from Dungeon Remove Spell) detail video

https://youtu.be/hkpUf3V_Puk


Scripts/Services/ExploringTheDeep/Items/QuestItems/StorageLocker.cs
```diff
@@ -21,7 +21,7 @@ public class StorageLocker : FillableContainer
 
         private bool m_Active;
         private Parts m_Type;
-        private List<Item> m_Barrels;
+        public List<Item> m_Barrels;
```
kevin-10: Standard practice is to make the private field of m_Barrels accessible through a property:

public List<Item> Barrels
{
    get { return m_Barrels; }
    set { m_Barrels = value; }
}

kevin-10: Standard practice is to make the private field of m_Barrels accessible through a property:

public List Barrels
{
    get { return m_Barrels; }
    set { m_Barrels = value; }
}

Tasanar:
So wipe the entire dungeon before adding this in?

Argalep:
spawns, teleporter, sparks and doors

kevin-10:
Is the delete old wrong called before the generate command is given?  I mean, the order doesn't really matter but it would be nice if it was called during generation.

kevin-10:
I will do some testing when I get home.  Thanks for being patient!

Issue #1859, opened 2017-03-16
Trashing blessed Items


Crome696:
EA-User reported that blessed Items (such as items) are able to be trashed on EA but not on ServUO.

kevin-10:
They can be trashed if they are on the list for turn-in points.  If not, you get a message.

Crome696:
I think what the player wanted to express is that any blessed item should be trashable.

PR #1865, opened 2017-03-18
Trashing blessed Items #1859 fix.

ruaduck:
You can delete blessed items per EA.  Tested with Spell Books, Fountain of Life Deed, and Robes that are blessed. Tested by ruaduck.


Tasanar:
You can ONLY trash blessed items that are worth clean up points. Try to trash any veteran reward, or bless a crimson cincture and try to trash it and it will not let you.

Tasanar:
This commit reverts a change Dexter added in months ago to address this. It's not perfect but I would rather have it the old way until all clean up items are added for points.

Tasanar:
I think more testing needs to be done. It will not let me trash a lot of blessed items in Test Center.

ruaduck:
Closed this pull request in light of Tasanar's and Kevin-10's comments.

PR #1867, opened 2017-03-20
BaseAI.cs AcquireFocusMob combatant acquisition optimization

GriffonSpade:
Split Aggressive AI and Uncontrolled Summoned Target/Passive AI/Special AI: Removed redundant checks from Aggressive AI (Closest, Strongest, Weakest). Those are all properly handled in BaseCreature.IsEnemy now.
Changed Uncontrolled Summon Target to check for null GetMaster rather than non-player, as some might conceivably have creature-summons on separate teams, which should attach each other.
Added note that faction/ethic enemy check only checks players and their pets, not creature with such faction allegiance.
Updated commented XMLSpawner faction check to use the standard code style.

### DIFF
--- a/Scripts/Mobiles/AI/BaseAI.cs
+++ b/Scripts/Mobiles/AI/BaseAI.cs
@@ -2905,91 +2905,84 @@ namespace Server.Mobiles
 							{
 								continue;
 							}
-
-							//Ignore anyone under EtherealVoyage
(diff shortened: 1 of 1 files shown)


kevin-10:
Works well.

kevin-10:
Once your PR is accepted, you can merge it.

PR #1874, opened 2017-03-22
fixs

Argalep:
* Prison Rat stat fix.
* Wrong Bedroll drop soap fix.
* Volcano message only player
* Demonic Jailor Speed Fix.
* Lockpick no time


kevin-10:
Why is the delay from lock picking removed?

Argalep:
No delay time for osi

https://youtu.be/hkpUf3V_Puk

kevin-10:
There appears to be a very very slight delay.  Make it 200 milliseconds, that way there is still justification for a timer.

PR #1876, opened 2017-03-23
Myrmidex Battle Spawner tweeks

kevin-10:
- fix to Myrmidex battle spawner routes, as previously pointed out by Argalep.

https://www.servuo.com/threads/myrmidex-spawn-problem.6776/

- Removed redundant FightMode.IsEnemy, Thanks to GriffonSpade for making the AcquireFocusMob in BaseAI.cs manageable!
- Fixed IsHostile and removed redundant FightMode.Aggressor check
- Added New Wrong to CreateWorld Gump


GriffonSpade:
Checked out the changes in BaseAI, and noticed a potential problem revealed, at least for modders. (Not a bug): Creatures will continue trying to fight an invalid target, rather than trying to find a new one.

Example, a Frenzied Ostard who is attacking me as a player will continue to stare at me after I've used staff orb to become an admin, and thus Blessed, despite being attacked by another mobile.

It may not be desirable to have them instantly drop target, though. I propose to add in the various AI's' 'DoActionCombat' modules, specifically to the 'in perceptionn range' check, if the target is Blessed.

For example, line 49 of MeleeAI.txt is:
            if (!m_Mobile.InRange(combatant, m_Mobile.RangePerception))
and would become:
            if (!m_Mobile.InRange(combatant, m_Mobile.RangePerception) || (combatant is Mobile && ((Mobile)combatant).Blessed))

This way, they will attempt to change targets as normal. May be instant, or take up to 10 seconds by default. At the same time, they'll not forget their target if it's just a temporary invulnerability ability.

kevin-10:
This was an issue long before the changes to AcquireFocusMob were made.  We can start a thread in the forums, but yea, I've noticed this too.

GriffonSpade:
On L52 of BaseCreature, after you delete the FightMode Enemy, shouldn't the comma after Good on the previous line be removed?

kevin-10:
Unless there is some reason I don't know about, I've seen it either way, both of which don't seem to make a difference.

PR #1881, opened 2017-03-24
Generalize MilitiaFighter check in IsEnemy

GriffonSpade:
-Generelized MilitiaFighter check in BaseCreature.txt IsEnemy
--Added virtual bool IsMilitiaFighter (default: false) to BaseCreature.txt
--Override bool IsMilitiaFighter to true in MilitiaFighter.cs
Allows any BaseCreature type to be set to IsMilitiaFighter, rather than this check being specific to only one creature type.

-Cleaned IsFriend in BaseCreature.txt
-Cleaned IsEnemy in BaseCreature.txt

Do not accept until after #1890 Bug Fixes, etc pull.


kevin-10:
Fix the conflict and its good.

GriffonSpade:
Heh, conflict was just removing that extraneous comma at the end of the FightMode enum after Good. Ready to roll.

PR #1886, opened 2017-03-26
Basic logging microframework with colored console logging support.

ppardalj:
This pull request introduces a basic framework for logging, with interface similar to log4net, which provides typical log levels.

## Usage

To use a log in your class, just add this line to the class code:
```
private static readonly ILog log = LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
```

Then just use the logger in your code instead of calling `Console.WriteLine` directly:
```
Console.WriteLine("Warning: Oops! You may look here for bugs!");
```
now becomes:
```
log.Warning("Oops! You may look here for bugs!");
```

## Background

At the moment, the provided implementation just outputs the messages to stdout, along with the category (which is simply taken from the name of the calling class), with a color depending on the log level.

I think this is good to standarize levels of message outputs and to decouple client code from log streams and settings, such as log category, formatting, etc.

In the future it can be used to provide alternative handlers for certain log categories which are now manually logged to file, such as network errors, firewall violations, etc. without need of writing custom code.

I've not added a dependency to log4net since the interface is pretty basic, but depending on your feedback it could be added as well to implement missing logging features.


dmurphy22:
Amazing contribution. This will take a little longer than normal to review due to the sheer size of it though. Please bare with us.

ppardalj:
@dmurphy22 no worries, no need to rush :) if I can support in any way, like explaining some parts, just let me know.

ppardalj:
What is missing for this to be merged? @Vorspire @dmurphy22

Issue #778, opened 2016-04-30
Staff related "bugs"


BugBot:
So today i installed ServUO and i noticed a couple bugs that i could have opened different threads for but, lazy. I've found some bugs that i thought are not right, concerning staff members.

1) A paragon greater mongbat aggressed me on ilshenar for no reason as i was walking around a champion spawn, as my owner player, this only seem to happen with paragons so the search field is not wide.

2) When i set myself as a murderer on my staff account, i can only see the felucca entry, but wouldn't it be more convenient to allow the staff members to see the "disabled" options? such as a red Ilshenar, but giving them the ability to actually travel there?
It's not much and it can easily be coded without being official, and it's sort of counter intuitive to be a murderer as a staff member, but it could be useful at some point idk.

3) When a staff member kills a player, he can get reported for murder, which should probably not be allowed, as the staff members are there to manage the shard and surely have the right to kill a player without a word. See #2 concerning travel disallow.

4) When i started, i gave myself heavy equipments with my bare 10 strength, and i was able to take them on, but, that is not the problem;
- When i attempted to equip a weapon (wild staff), i was not able, while there was no problem with armors.
- When i set myself some higher strength [11], everything falls in my backpack, which is correct. The problem is if i set my strength to 125, only my circlet falls in my backpack, probably due to the elven only property.
- With an elven only item [a circlet], when i set an attribute, it falls in my backpack.

---

So to sum up, this does not look uniform and in my opinion staff members should not be subject to restrictions, if they should abide to these, it should atleast be uniform.

So yea, that's it for now.
If i ever get some time later i might check if i can fix them myself and put a fix on git if wanted.

---

Sidenotes:
- I did not check if those bugs were already reported, as i am lazy, sorry if this is not correct.
- Since it is staff-side, i doubt anyone would be able to validate if it happens on OSI, so i guess it could be flagged as improvement/feature instead of bug.
- I was on my shard, when i was going to report the bugs, i switched to my browser and i was already in the bug report screen. That's what i call service. xD

Reported By: Po0ka
Report Thread: http://servuo.com/threads/staff-related-bugs.4637/

PR #1891, opened 2017-03-27
PublicMoongate.cs: Staff Overrides

GriffonSpade:
-Allows staff to use moongate regardless of status or range.
-Removed all occurances of 'this.' in the code.
-Replaced space indents with tab indents
-Added braces to all checks

Part of #778 Staff related "bugs" [https://www.servuo.com/threads/staff-related-bugs.4637/](url)

### DIFF
--- a/Scripts/Items/Functional/PublicMoongate.cs
+++ b/Scripts/Items/Functional/PublicMoongate.cs
@@ -80,7 +80,7 @@ namespace Server.Items
 				return;
 			}
 
-			if (from.InRange(GetWorldLocation(), 1) || from.AccessLevel >= AccessLevel.Counselor)
+			if (from.InRange(GetWorldLocation(), 1) || from.IsStaff())
 			{
 				UseGate(from);
 			}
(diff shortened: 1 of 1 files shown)


kevin-10:
Whats the point of that editor file?  I thought it auto-formatted?

GriffonSpade:
Might be related to the autoformatting -- I use Notepad++ to edit. 
I suppose I can convert them easily enough, though working with space indents is a bit of a pain, and I notice they tend to deform into irregular 3 and 5 space indents in places and need to be cleaned.

Give the word, boss, and I'll make it so.

GriffonSpade:
Alrighty, I got that plugin, that should help. But ah, where do I edit the editorconfig file for it? I don't see anything about that at all.
(And I was calling @kevin-10 boss ;))

ppardalj:
@GriffonSpade don't play with my heart like that :(

Hahah now seriously, I'm happy your PR is finally merged, congratulations!! :)

PR #1894, opened 2017-03-28
Removed compilation error in base creatures

techy2493:
Removed IsEnemy on line 7455.

This made the code unable to compile.


kevin-10:
Thank you.

PR #1898, opened 2017-03-28
Familiar AI Fix/Revamp v3

GriffonSpade:
Associated Thread: [Familiar AI Fix/Revamp](https://www.servuo.com/threads/familiar-ai-fix-revamp.6961/)

Before being merged to master, this pull should definitely be tested, preferably have confirmation in OSI behavior fidelity, and any necessary refinement to correct if either of the previous two have issues.

-Familiars ONLY attack creatures within RangeHome (default 10) of master.

v2:
-Fixed a a bug causing familiar to not attack anything not attacking itself
-Various incidental Trims

v3:

### DIFF
--- a/Scripts/Mobiles/AI/BaseAI.cs
+++ b/Scripts/Mobiles/AI/BaseAI.cs
@@ -581,7 +581,7 @@ namespace Server.Mobiles
 
 									if (m_Mobile.CheckControlChance(e.Mobile))
 									{
-										
+
 										m_Mobile.ControlOrder = OrderType.Guard;
                                         m_Mobile.ControlTarget = null;
 									}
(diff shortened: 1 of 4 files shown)


kevin-10:
Upon aggroing, familiars are auto-attacking the target.  They should not do this except death adder and vampire bat.

GriffonSpade:
I believe that was not my error, but the one who did that edit: AttacksMastersTarget was set to true in BaseFamiliar (I did not change it). I thought it was odd, but I didn't want to change any of the existing functionality.

GriffonSpade:
Reverted and changed some of the previous stuff. I think I might have this power-sliding and warping issue buggered. At least for familiars.

As a final question, will the familiars on OSI without AttackMastersTarget attack the enemy if they're the master's target AND happen to be in melee range? (essentially just not chasing them)

I want to go over version 3 a few more times, and I'll put it up in a new PR tomorrow.

GriffonSpade:
Updated this to version 3.

PR #1902, opened 2017-03-29
Adds a code standards document.

jdphenix:
This first iteration of a code standard is definitely not intended to be authoritative or final. My intent is to merely get the ball rolling on the idea of having code standards and to gauge interest on the idea.

Most of what I've added here is based on what seems to be in the majority of the ServUO code base. In other cases, it's just what I see as common sense practices (like descriptive names). 

I was to reiterate that I'm not trying to impose my ideas of standard on a project I've never contributed to, but rather to introduce the idea of having coding standards, which we could turn around and evolve into build dependencies on new code. 

I would not consider this pull request "final" right now.

### DIFF
--- a/CodeStandards.md
+++ b/CodeStandards.md
@@ -6,10 +6,10 @@
 
 ### Omit `this` for member access. 
 
-Do not use `this` to denote access to an instance's members. If `this` must be used to to access a
+Do not use `this` to denote access to an object's members. If `this` must be used to access a
 member, rename or refactor the code to not require it. 
 
-**Exception**: Scripts may use `this` to denote member access.
+**Note**: A solitary `this` may be used to reference the current object itself. 
 
 ### Use immutable data when possible. 
 
@@ -92,7 +92,7 @@ For example,
     public static bool Enabled = true; 
     public int TotalGold { get; }
 
-### Method parameters must be in pascalCase.
+### Method parameters and local variables must be in pascalCase.
 
 For example, 
 
--- a/CodeStandards.md
+++ b/CodeStandards.md
@@ -4,6 +4,35 @@
 
 ### Use tabs for indents. 
 
+### Add contrite, descriptive comments above code that is complex, counter-intuitive, or has exceptions to standards that are required to function.
+
+    // Example: you have a faction check, but it only checks players and their pets, ignoring uncontrolled faction creatures.
+
+    if (IsStaff())
+    {
+        //Staff can always do this!
+    }
+    else if(Criminal || Kills >= 5)
+    {
+        return;
+    }
+
+    // bad example - such a construct would require an explanation
+
+    if (Criminal || Kills >= 5 ? "T" : "F" == "T") 
+    {
+        return; 
+    }
+
+    // bad example - doesn't need explanation.
+
+    // if the mobile is a criminal
+    if (Criminal)
+    {
+         // reduce their health by 10.
+        Health -= 10;
+    }
+
 ### Omit `this` for member access. 
 
 Do not use `this` to denote access to an object's members. If `this` must be used to access a
--- a/CodeStandards.md
+++ b/CodeStandards.md
@@ -1,0 +1,139 @@
+# ServUO Coding Standards
+
+## General Conventions
+
+### Use tabs for indents. 
+
+### Omit `this` for member access. 
+
+Do not use `this` to denote access to an instance's members. If `this` must be used to to access a
+member, rename or refactor the code to not require it. 
+
+**Exception**: Scripts may use `this` to denote member access.
+
+### Use immutable data when possible. 
+
+When possible, use `readonly` and private setters on properties to enforce immutability. 
+
+### Do not use magic numbers or values. 
+
+Name values appropriately to what they mean. Do not use names for very obvious meanings. 
+
+For example, 
+
+    private readonly int m_MageryCircles = 8;
+
+    if (m_MageryCircles < castingSpell) 
+    { 
+        // something went wrong!
+    }
+    
+    // bad examples
+    if (8 < castingSpell)
+    { 
+        // something went wrong!
+    }
+
+    private readonly int m_one = 1; // unneeded clarification
+
+### Do not add `using` statements that aren't used by the source file. 
+
+Do not add unused `using` statements to the header of source files. 
+
+## Access Control Conventions
+
+### Fields must not be exposed as `public`, `protected`, or `internal`.
+
+For example, 
+
+    private int m_Gold; 
+
+    // bad examples
+    protected int m_Gold; 
+    public int m_Gold;
+
+## Naming Conventions
+
+### Prefer descriptive names.
+
+Use descriptive names in the ServUO code base. Use your best judgement for length. Be kind to 
+yourself three months from now.
+
+For example, 
+
+    int goldGained;
+    bool isLlamaSafe;
+
+    // bad examples
+    int gg;
+    int goldGainedTodayPerCharacter;
+    bool lsafe;
+
+### Private members must be prefixed with `m_` in CamelCase. 
+
+This rule applies to both instance and static members, including events, fields, properties, and 
+methods.
+
+For example, 
+
+    private bool m_CanReportMurder;
+    private static m_ParticleSupportType = ParticleSupportType.Detect;
+
+    // bad examples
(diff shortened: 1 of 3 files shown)


GriffonSpade:
A few things I noticed:
Please forgive me if I butcher the lingo.

1) local variables don't use 'normal' camel-case. They always start with a lower-case, then camel-case thereafter.
ex. myVariable, target, combatant, maxCircle, master, mc, etc.
I assume that this is so they don't interfere with Public variables.

2)
> +### Omit `this` for member access. 
> +
> +Do not use `this` to denote access to an instance's members. If `this` must be used to to access a
> +member, rename or refactor the code to not require it. 
> +
> +**Exception**: Scripts may use `this` to denote member access.

This seems garbled. First line says 'omit this for member access', last line says 'scripts may use this to denote member access'. 

I'm thinking this should maybe read: 
Do not prepend 'this.' to the current object's variables, functions, etc.
**Note** Solitary 'this' may be used to denote the current object itself

3) Comments.
Please add contrite, descriptive comments above code that is complex, counter-intuitive, or has notable exceptions to 

example: you have a faction check, but it only checks players and their pets, ignoring uncontrolled faction creatures.

-If you have a check that is completely empty, please place a contrite comment inside it that explains why it's there. Example:
if ( IsStaff())
{
//Staff can always do this!
}
else if{ Criminal || Kills >= 5 )
{
return;
}

4) Variables/functions in Abstract classes or otherwise intended to be used by objects/classes that inherit them, but not designed to be accessible in properties should also be Public, yes?

CodeStandards.md
```diff
@@ -0,0 +1,139 @@
+# ServUO Coding Standards
+
+## General Conventions
+
+### Use tabs for indents. 
+
+### Omit `this` for member access. 
+
+Do not use `this` to denote access to an instance's members. If `this` must be used to to access a
```
ppardalj: Typo: "to to" should be "to".

CodeStandards.md
```diff
@@ -0,0 +1,139 @@
+# ServUO Coding Standards
+
+## General Conventions
+
+### Use tabs for indents.
```
ppardalj: At the moment current `.editorconfig` states spaces are the current standard. I'm not sure whether the tab vs spaces discussion has ever happened in the past, so maybe project maintainers can provide some insight here.

In any case I would try to keep coding standards compliant with editor settings.

kevin-10:
I don't know how its done in the 'coding' world, but it just seems to me, like anything else, is that everyone has their own style.  Don't we all use some type of editor where it normalizes the indents?  I use Visual Studio, so the only time I see a discrepancy in indention is when you read them straight from github on my web browser.  Now, I see a lot of fancy words for things like camel case, something in which I didn't even know there was a name for!  I do this only for a hobby, so all this code standards stuff is new to me.  I mean, if the code works, and is efficient, who gives a sh*t which letters I capitalized.  If I want to name a field _HiThere instead of the trendy m_HiThere, what does it matter?  The RunUO, and now the ServUO environment is filled with different coding styles.  Bottom line, you either understand C# or not.  If you understand it, then this shouldn't be an issue.

GriffonSpade:
It's called CamelCase because ItHasHumpsInsteadOfSpaces. :)

Nope, Notepad++ doesn't automatically normalize the indents on save, it saves the file exactly as you leave it. 
Basically, tabs are more visible(nice long line instead of tiny dots), require dealing with less characters overall, and no broken indents because one character is one indent step. 
(Even with the editorconfig plugin, I wound up just using tabs anyway and converting to spaces manually. Mind, not on the Familiar AI stuff. Don't want to distract from what's actually being changed with a billion style edits.)

Basically, because each different type of variable has its own standardized format, it makes everything intuitive to read and use for everyone involved. It can be extremely confusing and time-consuming to read/edit someone's work who doesn't use a standardized coding style. (How confusing and time-consuming depends on how deviant it is.) And starting variables with underscores makes people's heads explode.

kevin-10:
I see your point, but imo it's nit pickey. But I don't do this for a living, and I suspect some of you do so I'm assuming any type of standardized code formatting is 101 stuff.  I'll do my best to adhere to the guidelines from here on out.  For the most part, I'd say I'm pretty good about most of this stuff.  I do like starting fields with underscore lol.

GriffonSpade:
You're right that it's nitpicky, and always annoying. But at the same time, when you're looking over someone else's work, you're so very thankful it's there.

CodeStandards.md
jdphenix: I think it would be reasonable to base off the `.editorconfig` that's present at the moment - including updating it and the ReSharper settings file that also in the repo with what (if anything) we end up deciding here.

jdphenix:
@kevin-10 Keep in mind I'm not saying "thou shalt" in this suggestion. I'm frankly not the right person to make the decisions made in my PR. I actually personally prefer spaces (flame on!), but tabs are what's in most of the source files. I'm just trying to start the discussion. 

What I did was just base mostly off of what is present in the repo right now, which doesn't necessarily reflect what ServUO's primary maintainers want to do today. 

I definitely wouldn't call standards strictly required, simply a decent idea in my opinion.

ppardalj:
@kevin-10  Coding standards are a basic thing for every serious software project, even more important for those with a large codebase such as ServUO's. Especially now, when the development is focusing more in quality over quantity, it's not only important that code works, but that it's readable by other contributors, and it's easily modifiable. I recommend to read about the topic if you're interested: https://en.wikipedia.org/wiki/Coding_conventions

Even in the open source world, where people don't develop for profit, this is as important (if not more) than in proprietary code, given the potential amount of contributors. See some examples here https://balau82.wordpress.com/2010/12/28/coding-styles-comparison-in-the-open-source-software-world/

Personally, I do software engineering for a living, and I can't work with a codebase that does not follow any type of standards, it's just too much time consuming. I'm sure every software engineer that has ever contribute to ServUO agrees that it's sometimes difficult to read code in this project, and that they will be encouraged and more motivated to contribute if they see a clean codebase.

For less experienced people, it's also a huge benefit since they can use the project codebase to learn good programming practices and boost their learning speed, which at the end of the day also benefits the project itself by improving the quality of their contributions.

kevin-10:
Understood.  I learned c# in the RunUO/ServUO environment, so I feel like the standards are ok.

jdphenix:
One other thing I need to ask about 

To `var` or not to `var`, that is the question @ppardalj @kevin-10

GriffonSpade:
I think it would require some consensus from the majority senior developers on the tab vs space issue.

As to if it's been discussed before, I linked it at the bottom of your general ServUO questions thread and I'll put it here too, my thread from back in November:

[https://www.servuo.com/threads/creature-file-formal-style.5642/](https://www.servuo.com/threads/creature-file-formal-style.5642/)

ppardalj:
@jdphenix regarding your questions:
- Tabs vs spaces: as you pointed, it does not really matter as long as there is consistency across the whole codebase. If the majority of the code is already formatted with tabs, I'd go for tabs.
- `Var` usage: I would leave this up to the developer, use `var` only if it enforces readability of the code. For example, for long initializers such as `Dictionary<Mobile, int> damageEntries = new Dictionary<Mobile, int>()` it makes sense to use `var` to remove duplication. This assumes we are using good naming for variables and that type can be omitted without reducing readability. This article summarizes it pretty well: http://intellitect.com/when-to-use-and-not-use-var-in-c/

GriffonSpade:
Actually, I discovered another reason to use tab indents: Space indents don't work in chat here or on forums, while tab indents do :P

GriffonSpade:
It's important to note that line 11-18 is a separate example from line 9. Needs another // Example: line inserted.

GriffonSpade:
Space Indents apparently cause issues with configuration files. Is there any way to exclude specific file types? Otherwise, shall we just go over to tabs?

CodeStandards.md
Vorspire: Tab spacing always comes out nicer.

Vorspire:
There is never going to be a unanimous agreement between everyone on the standards that should be used throughout the project. Coding standards change with time and for that reason it's best to stick to resources that are actually going to stay up-to-date. This information would be better suited for the ServUO forum for those that haven't already adopted coding standards or developed their own personal style.

Issue #1712, opened 2017-02-12
Missing feature - world chat


BugBot:
World chat has been something real UO has had for a while.

Hitting   ,  allows you to chat on real UO.

I know we have plenty of great chat programs on real UO but since hitting , already brings up a chat interface could we get this tied into it?

A chat system with channels and the ability to block accounts from access.

This is all pictures from real UO Atlantic.

Hitting CHAT brings up this menu - General, Help, Trade, and Looking For Group, are the stock. The other ones are ones players made. General is the default (unless on a trial account) and most used.

[ATTACH=full]8163[/ATTACH]

This is the default color and how it looks in world and journal

[ATTACH=full]8164[/ATTACH]

I narrowed it down to exactly this color on the dye tub

[ATTACH=full]8165[/ATTACH]

Reported By: Tasanar
Report Thread: http://servuo.com/threads/missing-feature-world-chat.6632/

GriffonSpade:
Daughter Issue for Chat Moderator controls: [https://github.com/ServUO/ServUO/issues/1913](url)

ppardalj:
@Ravenwolfe @kevin-10 this can be closed as https://github.com/ServUO/ServUO/pull/1904 was already merged.

PR #1904, opened 2017-03-29
World chat

ppardalj:
## Summary

This pull request adds the missing feature **world chat**, as requested in https://github.com/ServUO/ServUO/issues/1712. It's basically a rework of the legacy chat system, which removes all the functionality which is no longer supported by the client (room moderators, nick change, voice, etc.) and ensures compatibility with the latest client versions.

By default the global chat is enabled, players are allowed to create channels and four default channels are created, as per OSI: "Help", "General", "Trade" and "Looking for group".

Users are no longer able to select their nick, their character name is always shown instead.

## Configuration

A configuration file is added to modify settings: `Config/Chat.cfg`.

A config setting is added `AllowCreateChannels` is added to control whether players are allowed to create new channels, otherwise they are restricted to the statically defined channels.

## Logging

A logging framework is also introduced to log every chat message to log files under the `Logs/Chat` directory, for auditing purposes. Chat message and channel events (creation, deletion, join, left) are logged globally and per-channel basis.

## TODO

- Configure default channels using XML configuration rather than hardcoded.
- Capabilities for staff to manage chat.


GriffonSpade:
So far, haven't been able to get chat to do anything. No channels appear, even when trying to create them, etc. Is there something special that must be done in-game to activate it?

ppardalj:
@GriffonSpade thanks for the report, I forgot to commit needed changes in the core, my mistake ^^u it should work now :)

ppardalj:
@kevin-10 @GriffonSpade thanks for the feedback!

Regarding the hue, it's actually defined client side; every command 0xB2 is rendered with the same color (and I've not actually found how to change it client side.)

As far as I could test with my former OSI account in the past, there's no more chat functionalities players can access other than what's implemented here, but I'm pretty sure GM's should have admin capabilities, which in ServUO should be implemented as custom content, of course. Any contribution in this regard is welcomed :)

GriffonSpade:
Found it. It's manually controlled in the client. It's in the eyeball tab with speech, emote, party, guild, and alliance colors.

So we can consider that a non-issue 👍

PR #1906, opened 2017-03-30
Fixs 2

Argalep:
* add Nexus in Exodus Dungeon
* add Nexus Deed
* Fezzik fix

[GenExodusNexus


kevin-10:
Also, thanks for taking out the 3 tiered cake.  I see you noticed it in the new BOD publish.  That'll save me some time squashing conflicts.

PR #1908, opened 2017-03-30
Replacing OppositionGroup with OppositionList v1.1

GriffonSpade:
Alright guys, this is ready, and should work better. However, it's replacing stable, if clunky, code, so please test and review for a while to make sure nothing unexpected becomes a problem.

-Simpler, more dynamic, more elegant than OppositionGroup.
-OppositionGroup has an XvsY-group virtual enum, along with two lists of types, each of which are checked until it finds the type in the list.
-OppositionList has an X-group virtual enum, which has two switch(), first for its own OppositionType to pick the correct function, then a second for the target's OppositionType which determines if it has an enemy OppositionType.
-The first few checks are nearly identical to OppositionGroup(modified for OppositionType), making sure 'this' has an OppositionType, then checking if the target is a mobile, then checking if the target has an OppositionType. (A creature having a ControlMaster is ignored, just as in OppositionGroup)
-entirely handled within BaseCreature.cs (with the OppositionType set within the creature's type file)
--No need to update a clunky list for each new member of the Opposition, simply add the enum into its creature file
-OppositionGroup hasn't been completely excised for compatibility reasons
-Easily expandable, including the ability to add more than a single enemy group
-All mobiles with currently functioning OppositionGroup converted to OppositionList
-Savage Shaman is Female-Only
-Savage Rider is Male-Only

### DIFF
--- a/Scripts/Mobiles/Bosses/LordOaks.cs
+++ b/Scripts/Mobiles/Bosses/LordOaks.cs
@@ -8,43 +8,44 @@ namespace Server.Mobiles
     {
         private Mobile m_Queen;
         private bool m_SpawnedQueen;
+
         [Constructable]
         public LordOaks()
             : base(AIType.AI_Mage, FightMode.Evil)
         {
-            this.Body = 175;
-            this.Name = "Lord Oaks";
+            Body = 175;
+            Name = "Lord Oaks";
 
-            this.SetStr(403, 850);
-            this.SetDex(101, 150);
-            this.SetInt(503, 800);
+            SetStr(403, 850);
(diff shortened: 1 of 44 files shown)


kevin-10:
What about Gray Goblin vs. Green Goblin?

GriffonSpade:
It's not using OppositionGroup functions (If it's even working). Really, OppositionGroup is a mess right now. It's not been kept up-to-date, and there are all sorts of things missing from it.

I want to make sure the foundation is working properly before adding a bunch of additional fixes into it. And it should make it FAR easier to keep updated, as well.

PR #1911, opened 2017-04-01
Made it so that ServUO can be launched without a console attached.

techy2493:
Added a try/catch to catch errors on main.cs when launching the server without a connected console.

Also changed how input is listend for on owner account, you now have to
press enter but it not longer requires a console be connected to the
server directly.

### DIFF
--- a/Scripts/Misc/AccountPrompt.cs
+++ b/Scripts/Misc/AccountPrompt.cs
@@ -12,7 +12,9 @@ namespace Server.Misc
                 Console.WriteLine("This server has no accounts.");
                 Console.Write("Do you want to create the owner account now? (y/n)");
 
-                if (Console.ReadKey(true).Key == ConsoleKey.Y)
+                string Key = Console.ReadLine();
+ 
+                if (Key.ToUpper() == "Y")
                 {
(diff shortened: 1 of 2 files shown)


techy2493:
You are probably right. It's what I get for writing code too late at night. I will drop this PR and look at a better approach.

techy2493:
The only problem I have with your suggestion is that user input is not the only way to send input to the console. 

I'm going to leave anywhere that actually requires user input on core.service but change it to console.readline instead of readkey which causes an exception if there is no attached console as I did in this PR. 

This will allow a control service to send input to the system if necessary but not use anything that requires an actual system console. 

@ppardalj  I will tag you in the updated PR.

ppardalj:
@techy2493 what other ways are there for providing input? I mean, for a regular process for sure there are ways such as piping commands or redirecting the standard descriptors, but at the moment ServUO does not use any of these as far as I know. Do you have something in mind? I'd love to hear more :)

Have you tried `Environment.UserInteractive` without an attached console? Can you tell us what is the result? Afaik it could be used in this case to prevent any exception when using `Console.ReadKey`, but now you made me doubt :p

PR #1917, opened 2017-04-01
Allow ServUO be operated interactively without a console.

techy2493:
This PR allows ServUO to operate in interactive mode without an active console. 
This is useful for server admins who would like to programattically control the IO streams of ServUO.

Adds Core.NoConsole and -noconsole command line argument.

Addressed issues raised by @ppardalj


Server/Main.cs
```diff
@@ -385,8 +387,17 @@ public static void Main(string[] args)
 				{
 					_UseHRT = true;
 				}
+                else if (Insensitive.Equals(a, "-noconsole"))
+                {
+                    NoConsole = false;
```
ppardalj: Did you mean `NoConsole = true`? (also below)

Server/Main.cs
```diff
@@ -523,7 +541,7 @@ public static void Main(string[] args)
 
 				Console.WriteLine(" - Press return to exit, or R to try again.");
 
-				if (Console.ReadKey(true).Key != ConsoleKey.R)
+				if (Insensitive.Equals(Console.ReadLine(), "R"))
```
ppardalj: IMHO no need to change this, there is a check for `Core.Service` two lines above. Just change it for `Core.NoConsole` should be enough.

Scripts/Misc/AccountPrompt.cs
```diff
@@ -12,7 +12,9 @@ public static void Initialize()
                 Console.WriteLine("This server has no accounts.");
                 Console.Write("Do you want to create the owner account now? (y/n)");
 
-                if (Console.ReadKey(true).Key == ConsoleKey.Y)
+                string Key = Console.ReadLine();
```
ppardalj: Local variables should be camelCased (i.e. prefer `key` over `Key`).

Scripts/Commands/ConvertPlayers.cs
```diff
@@ -51,9 +51,9 @@ public static void Convert_OnCommand(CommandEventArgs e)
                 NetState.ProcessDisposedQueue();
                 World.Save();
             
-                Console.WriteLine("{0} players have been converted to PlayerMobile. {1}.", count, Core.Service ? "The server is now restarting" : "Press any key to restart the server");
+                Console.WriteLine("{0} players have been converted to PlayerMobile. {1}.", count, Core.NoConsole ? "The server is now restarting" : "Press any key to restart the server");
                 
-                if (!Core.Service)
+                if (!Core.NoConsole && !Core.Service)
```
ppardalj: Why checking both? Is not enough to check just `Core.NoConsole`?

Server/Main.cs
```diff
@@ -87,6 +87,8 @@ public static TimeSpan ProfileTime
 		}
 
 		public static bool Service { get; private set; }
+
+        public static bool NoConsole { get; private set; }
```
ppardalj: I would propose to change the name of this property to `NoInteraction`. IMHO it reflects better the intention of this property, which is to avoid querying the user through the console for input during execution.

Server/World.cs
```diff
@@ -288,7 +288,7 @@ public DataEntry(SaveData data, int typeID, string typeName, long pos, int lengt
 					{
 						Console.WriteLine("Error: Type '{0}' was not found. Delete all of those types? (y/n)", typeName);
 
-						if (Console.ReadKey(true).Key == ConsoleKey.Y)
+						if (Insensitive.Equals(Console.ReadLine(), "Y"))
```
ppardalj: Same here, you have a check of `Core.Service` right above. Just change it to use `Core.NoConsole` should be enough.

Server/Main.cs
techy2493: The goal of this PR is not to remove all user interaction. Core.Service and -service do this effectively. The primary goal of this PR is to allow the administrator to force the use of IO methods which do not require a console. 
This frees the administrator to attach to the IO streams of the ServUO application and supply input and read output with third party applications.

Server/Main.cs
techy2493: Ah. You are correct. I will update this PR.

Server/Main.cs
techy2493: This is user interaction which is desirable. We're not replacing Core.Service just extending it's functionality and making it so that a standard console tied to the IO streams is not needed.

Scripts/Commands/ConvertPlayers.cs
techy2493: No. Service and NoConsole have two different intended purposes. Please see my other replies.

Scripts/Misc/AccountPrompt.cs
techy2493: Sorry. Very true. I will update this PR.

Server/World.cs
techy2493: Again, no. Service and NoConsole have two different intended purposes. Please see my other replies.

techy2493:
Several replies have been to the effect of making NoConsole replace the functionality of Service.

I need to stress that these are for two entirely different use cases.

The first and existing is Core.Service. This should be used when ServUO is executed as a system service and daemon where no input from the user or otherwise is expected.

The second and new is Core.NoConsole. This should be used for when ServUO is executed as a child process of another application which binds and consumes it's IO streams. In this case input (user or automated) may or may not be available. In the former an application is reading and providing input to ServUO based on a new consoled presented to the user or automated responses to stimuli. In the latter, while the application is bound no input is expected in which case both NoConsole and Service should be set to true.

techy2493:
Also, @ppardalj  Noted a couple of issues. I will update them this afternoon. Please do not merge this until I comment they have been made.

Server/Main.cs
ppardalj: Ok, I think I misunderstood the purpose. Everything allright then 👍

Server/Main.cs
ppardalj: Got it :) Thanks for the clarification!

ppardalj:
@techy2493 thanks for the clarification about the use case, I misunderstood the purpose of the change. Now it makes sense to me. Sorry about the noise!

techy2493:
@ppardalj Haha, no thank you for being so observant and caring about what goes into the repo.

You were right on my first set of changes. They were bad code. I won't try to defend. This new implementation is much more robust and catches several places I missed before.

I've had some things arise I will push the discussed changes shortly.

techy2493:
I have addressed the new issues raised by @ppardalj .

This is ready for review / merging.

mholmes3038:
Looks awesome, thanks for sharing

techy2493:

Hi Guys, this has been sitting for a while. Can I get an idea of your thoughts on it, and what chance it has of getting approved?

Server/Main.cs
kevin-10: This is not working.  Insensitive.Equals(Console.ReadLine(), "R") is not reading the first r on the line.  You either have to put a capital R or you have to put more than one "r".

PR #1919, opened 2017-04-02
Goblin Opposition

GriffonSpade:
Associated Thread: [https://www.servuo.com/threads/opposition-green-goblin-vs-grey-goblin.6992](url)
-Added HasFoughtPlayer property.
--Available in props
--Defaults to false
--Set True in OnCombatantChanged() when Combatant is player or player pet
--Set True in AggressiveAction() when ControlMaster is null and aggressor is player or player pet
--Is not saved
--Reverts to false on sector deactivation
-Added GreenGoblin and GrayGoblin Opposition Types with HasFoughtPlayer check.
-Added HasFoughtPlayer check in Terathan and Ophidian Opposition checks.

-Added OppositionType.GrayGoblin to GrayGoblin, GrayGoblinKeeper, and GrayGoblinMage
-Added OppositionType.GreenGoblin to GreenGoblin, GreenGoblinAlchemist, and GreenGoblinScout

v2
-Deprecated Commented out HasFoughtPlayer functionality.

### DIFF
--- a/Scripts/Mobiles/Normal/BaseCreature.cs
+++ b/Scripts/Mobiles/Normal/BaseCreature.cs
@@ -994,6 +994,7 @@ namespace Server.Mobiles
         public virtual bool IsMilitiaFighter { get { return false; } }
 
         // Opposition List stuff
+/*
         private bool m_HasFoughtPlayer;
 
         [CommandProperty(AccessLevel.GameMaster)]
@@ -1006,7 +1007,7 @@ namespace Server.Mobiles
                 InvalidateProperties();
             }
         }
-
+*/
         public virtual OppositionType OppositionList{ get{ return OppositionType.None ; } } // What opposition list am I in?
 
         public enum OppositionType
@@ -1055,12 +1056,13 @@ namespace Server.Mobiles
 
         private bool m_TerathanEnemy(OppositionType egroup)
         {
+/*
             // Don't fight monsters after fighting players, until sector deactivate/server restart
             if (HasFoughtPlayer)
             {
                 return false;
             }
-
+*/
             switch (egroup)
             {
                 case OppositionType.Ophidian: return true;
@@ -1070,12 +1072,13 @@ namespace Server.Mobiles
 
         private bool m_OphidianEnemy(OppositionType egroup)
         {
+/*
             // Don't fight monsters after fighting players, until sector deactivate/server restart
             if (HasFoughtPlayer)
             {
                 return false;
             }
-
+*/
             switch (egroup)
             {
                 case OppositionType.Terathan: return true;
@@ -1121,12 +1124,13 @@ namespace Server.Mobiles
 
         private bool m_GrayGoblinEnemy(OppositionType egroup)
         {
+/*
             // Don't fight monsters after fighting players, until sector deactivate/server restart
             if (HasFoughtPlayer)
             {
                 return false;
             }
-
+*/
             switch (egroup)
             {
                 case OppositionType.GreenGoblin: return true;
@@ -1136,12 +1140,13 @@ namespace Server.Mobiles
 
         private bool m_GreenGoblinEnemy(OppositionType egroup)
         {
+/*
             // Don't fight monsters after fighting players, until sector deactivate/server restart
             if (HasFoughtPlayer)
             {
                 return false;
             }
-
+*/
             switch (egroup)
             {
                 case OppositionType.GrayGoblin: return true;
@@ -3982,12 +3987,13 @@ namespace Server.Mobiles
                     aggressor.Aggressors.Add(AggressorInfo.Create(this, aggressor, true));
                 }
             }
+/*
             else if (!HasFoughtPlayer && (aggressor.Player ||
                 (aggressor is BaseCreature && ((BaseCreature)aggressor).GetMaster() is PlayerMobile)))
             {
                 HasFoughtPlayer = true;
             }
-
+*/
             OrderType ct = m_ControlOrder;
 
             if (m_AI != null)
@@ -4327,12 +4333,13 @@ namespace Server.Mobiles
 
             if (Warmode)
             {
+/*
                 if (!HasFoughtPlayer && (Combatant is PlayerMobile ||
                     (Combatant is BaseCreature && ((BaseCreature)Combatant).GetMaster() is PlayerMobile)))
                 {
                     HasFoughtPlayer = true;
                 }
-
+*/
                 if (CanFly)
                 {
                     Flying = false;
@@ -7434,9 +7441,9 @@ namespace Server.Mobiles
             {
                 m_AI.Deactivate();
             }
-
+/*
             HasFoughtPlayer = false;
-
+*/
             base.OnSectorDeactivate();
         }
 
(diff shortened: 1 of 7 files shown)


kevin-10:
So, this issue where once a player is attacked, they won't attack other opposing monsters is on EA?  If this is true, this is an obvious EA bug in target acquisition.  I'm not a big fan of duplicating bugs for the sake of accuracy.

GriffonSpade:
What about making using this a configuration option? (defaulted to false, of course) It's how OSI has it, so we should probably have it possible... even if it's stupid.

Unfortunately, I have no idea how I'd implement that. Anyone informing me how I'd go about that would be appreciated.

kevin-10:
Checking opposition should be called in IsEnemy in BaseCreature, which is called during AcquireFocusMob, so essentially, it is handled in AcquireFocusMob.  If the creature already has a combatant, then AcquireFocusMob won't even be called so prioritizing is moot.  

Also, to answer a question regarding priority in, for example, evil creatures attacking positive karma creatures, this will take priority over any type of oppisition acquisition as it is called before IsEnemy in the method.

kevin-10:
And I'm testing the as is code in stock ServUO, and its working just fine.  Why fix something if its not broken?  Sure, we may need to update some creature scripts with the appropriate OppositionGroup override, but the mechanics involved work fine.

ppardalj:
@kevin-10 `AcquireFocusMob` is called every 10 seconds regardless of the combatant of the creature. This is how creatures sometimes change targets when you move around them, for example. The delay is defined in `BaseCreature` class:

```
/* To save on cpu usage, RunUO creatures only reacquire creatures under the following circumstances:
 *  - 10 seconds have elapsed since the last time it tried
 *  - The creature was attacked
 *  - Some creatures, like dragons, will reacquire when they see someone move
 * This functionality appears to be implemented on OSI as well
 */
public virtual TimeSpan ReacquireDelay { get { return TimeSpan.FromSeconds(10.0); } }
```

Regarding  where to put the logic, `IsEnemy` would also probably work, but IMHO it's not semantically correct, since creatures in opposition groups "are enemies" always. The decision to attack a certain enemy belongs in the algorithm to determine the focus mob, not the algorithm to determine whether a mob is an enemy or not.

The current ServUO is not working completely right in this regard. I was able to reproduce this, although you need to set the appropiate context. What I did was:
1. Add a therathan and an ophidian, set debug true.
2. Let they start attacking you.
3. You start running, so that they're closer between them than to you. 
4. Since they have `FightMode.Closest`, when they reacquire the target, they'll begin attacking themselves :)

Another different discussion is whether this is expected behaviour or not. Actually I'm in @kevin-10 side on this, IMHO the current mechanics are fine, I would leave it as is.

GriffonSpade:
FocusMob is a bad fit for a couple reasons. One is that it can be used for finding a friendly target. The other is that it is cleared every time the AI ticks, so it would have to be checked either all over the place or every single time the AI triggers.

AggressiveAction() was chosen because it encompases every instance of being attacked and OnCombatantChanged() because it encompasses every instance of attacking, covering both fields with no false positives or points of failure to my knowledge.

Further, IsEnemy is called in places other than AcquireFocusMob, for special abilities and AOE spells.

Thinking about it, I believe some of the code in AcquireFocusMob should be moved into IsEnemy so special abilities and AOE spells target properly. (Specifically subsuming lines 2908-2972 into IsEnemy)

GriffonSpade:
For reference, it isn't always called every 10 seconds. It's called every time it doesn't have a good combatant, and it aborts out of searching for a target on its own if 10 seconds have not elapsed since the last check.

kevin-10:
AcquireFocusMob is called from DoActionWander(), or when the focusmob/combatant is out of range or dead.  So, while the creature is actively fighting, AcquireFocusMob is never called.  The 10 second interval is simply returns out of the function, but it is still called if the above conditions are met.

I think its working fine.

ppardalj:
You're right with the `AcquireFocusMob` thing, my STR is probably due to some other reason, sorry about it!

GriffonSpade:
So apparently the underlying system was removed, which kinda screws this over.

What do you mean OppositionList v1.1 was 'incomplete'? It had everything functioning that OppositionGroup did.

If you expect me to have all the hundreds of mobiles researched, tested, and readied before implementing anything, then I quit.

kevin-10:
I had initially tested with goblins, and didn't realize goblin opposition was never added to the original system.  The new system, though works exactly like the old one, adds unnecessary code to reinvent the wheel.  All opposition checks was handled in OppositionGroup.cs, whereas the new system adds several methods to BaseCreature.cs.

GriffonSpade:
Do you want it all handled in a separate file? I could do it. It might take a little bit, but I could do it.

...and I'm not sure how you initially tested with goblins. Those still had orc vs savage oppositiongroup, and weren't included in the lists in oppositiongroup.cs

kevin-10:
It already is handled in a separate file, OppositionGroup.cs.

To edit the current/old system, you need to override a property in the creatures script, and add it to the list in the appropriate opposition group.  Thats not too bad!

GriffonSpade:
It looks like shortening it into a single method of the same length as the base method is easy enough.  It literally took under two minutes. Now where do we go from here?

GriffonSpade:
OppositionType ONLY needs the override property. No need to check a list.

GriffonSpade:
Specifically, all I had to do was delete the extra methods, and replace this switch:
[code]            switch (OppositionList)
            {
                case OppositionType.Terathan: return m_TerathanEnemy(c.OppositionList);
                case OppositionType.Ophidian: return m_OphidianEnemy(c.OppositionList);
                case OppositionType.Savage: return m_SavageEnemy(c.OppositionList);
                case OppositionType.Orc: return m_OrcEnemy(c.OppositionList);
                case OppositionType.Fey: return m_FeyEnemy(c.OppositionList);
                case OppositionType.Undead: return m_UndeadEnemy(c.OppositionList);
                default: return false;
            }
[/code]
with this switch:
[code]            switch (OppositionList)
            {
                case OppositionType.Terathan: return (c.OppositionList == OppositionType.Ophidian);
                case OppositionType.Ophidian: return (c.OppositionList == OppositionType.Terathan);
                case OppositionType.Savage: return (c.OppositionList == OppositionType.Orc);
                case OppositionType.Orc: return (c.OppositionList == OppositionType.Savage);
                case OppositionType.Fey: return (c.OppositionList == OppositionType.Undead);
                case OppositionType.Undead: return (c.OppositionList == OppositionType.Fey);
                default: return false;
            }
[/code]

Tasanar:
> "GriffonSpade commented 19 hours ago 
> What about making using this a configuration option? (defaulted to false, of course) It's how OSI has it, so we should probably have it possible... even if it's stupid.
> Unfortunately, I have no idea how I'd implement that. Anyone informing me how I'd go about that would be appreciated."

You guys have done a lot of the heavy lifting so I am not trying to nitpick. 

However in the past I have seen the main devs agree that IF we can do something better than EA, because it is obviously wrong, we should correct it. It makes us unique and in a lot of ways a better alternative.

We should not just add in stupid bugs or mechanics because it exists there. A lot of times if you actually look people have complained in length on stratics. Maybe real UO does not change it because their code is so entwined?


RunUO set this precedent when they corrected "smart house placement" years ago. 

Even in house placement today, on ServUO and RunUO, you can see an option to set it to OSI (stupid) or RunUO ("smart") house placement. 


If you guys can do something better PLEASE DO!!!

GriffonSpade:
@Tasanar does that mean we should make it a configuration like house placement, but have it set to 'smart way' by default, instead of 'stupid OSI way'?

GriffonSpade:
Closing this. It's defunct without OppositionList, and Opposition Tribes system obsoletes it.

PR #1921, opened 2017-04-02
Revert "Replacing OppositionGroup with OppositionList v1.1"

kevin-10:
Reverts ServUO/ServUO#1908

Reverting as it was incomplete, and more importantly, the old system worked just fine.

### DIFF
--- a/Scripts/Mobiles/Bosses/LordOaks.cs
+++ b/Scripts/Mobiles/Bosses/LordOaks.cs
@@ -8,44 +8,43 @@ namespace Server.Mobiles
     {
         private Mobile m_Queen;
         private bool m_SpawnedQueen;
-
         [Constructable]
         public LordOaks()
             : base(AIType.AI_Mage, FightMode.Evil)
         {
-            Body = 175;
-            Name = "Lord Oaks";
+            this.Body = 175;
+            this.Name = "Lord Oaks";
 
-            SetStr(403, 850);
-            SetDex(101, 150);
-            SetInt(503, 800);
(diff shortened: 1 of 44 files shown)


GriffonSpade:
What do you mean it was incomplete? It had everything that OppositionGroup had.

techy2493:

Can anyone explain how the code got in the repo in the first place if it was just going to be reverted in a few days? 

@kevin-10 You were vocal in the original PR, why did you not raise these concerns before?

kevin-10:
I was under the assumption that the original system didn't work by testing the wrong creatures.  Once I realized that, I went back to the old system and found it worked just fine.  With that being said, I apologize for any time wasted on this but the way I see it, it works just fine.  Let's not reinvent the wheel.  The  Not thing this system needs is the addition of gray vs green goblins.

GriffonSpade:
Um, I'm a bit confused how you came to that impression. This was the very first paragraph in the notes: "Alright guys, this is ready, and should work better. However, it's replacing stable, if clunky, code, so please test and review for a while to make sure nothing unexpected becomes a problem."

Further, there's a reason why we do not use wooden wagon wheels on cars. There's no point in reinventing the wheel, but there _**IS**_ a point in inventing a _**BETTER**_ wheel

kevin-10:
Cool, I get it.  If @Vorspire or @dmurphy22 decides to re-commit, I'm fine with it.  I believe the old system is much less clunky, and works exactly how its supposed to.  As for the merge and revert, I apologize.

PR #1922, opened 2017-04-03
Fix/duplicates - De-duplicates code in Serialization Server code

jdphenix:
Much of the serialization code in the Server project was duplicating itself. This PR resolves that. 

Save performance remains the same in my testing setup, which is admittedly only the results of `[CreateWorld`.


Server/IO/GenericWriter.cs
```diff
@@ -0,0 +1,248 @@
...
+		}
+
+        protected virtual void WriteCol<T>(IEnumerable<T> list, Action<T> writer)
+        {
+               foreach(var o in list)
+                   writer(o);
+        }
+        
+		public virtual void WriteItemList(ArrayList list, bool tidy = false) => 
+			WriteListTidy(list, () => WriteCol(list.OfType<Item>(), Write), tidy, i => ((Item) list[i]).Deleted);
```
Vorspire: There must be a better way to handle this, you're going to increase world save write time exponentially because each tidy check has to traverse the source list via [indexing].

The tidy predicate also does not check if the object is null and will crash.
A null object should be considered equivalent to a deleted object and thusly removed from the list.

jdphenix: I agree on the null check, will fix. If we don't end up merging in this PR, I'll put one in to fix this assumption in the current serialization methods (example: https://github.com/ServUO/ServUO/blob/master/Server/Serialization.cs#L725)

I disagree that save time increases exponentially with this change. With a fully populated world on my machine, save time remained constant. (You should have seen my first go around at this, it was comically slow) The current `WriteXXXList` methods all access the list with their index. Random access on a list is constant time. Same as above, I'm happy to explore better ways. 

My goal here was to make the minimum number of changes necessary while de-duplicating code. 

Additionally, I apologize for my slow response - GitHub didn't show a notification for your review.

Vorspire: You're right about the speed increase, I was improperly assuming that List[i] is O(n) and not O(1) - so we can disregard speed issues.

Then I guess it's all good!

As for the tidy callback, a simple Action<IEntity> might work better as IEntity implements the Deleted property and the lists that require tidying usually consist of IEntity-defined objects (with the exception of Guilds). Then you can reduce the need for casting and can still inline your null checks;

_Write( ... , obj => obj == null || obj.Deleted ) ); 

You can then supply list[i] as the argument to the tidy delegate, rather than the index itself.

But that's up to you hehe.

jdphenix:
https://github.com/ServUO/ServUO/pull/1922/commits/782a5a22023b28e64e882dc3da18102fd34dd71f

- Null checks added to predicates passed into `WriteListTidy` and `WriteSetTidy`. 
- Obsoleted Write methods accepting `ArrayList`. Only one core script depends on them. I'd say leave these for a few months and remove later.

PR #1924, opened 2017-04-04
Replacing OppositionGroup with Opposition Tribe v1.0

GriffonSpade:
Please test and give feedback.
[Associated Thread: Replacing OppositionGroup with Opposition Tribes](https://www.servuo.com/threads/replacing-oppositiongroup-with-opposition-tribes.7002/)
[Source Branch](https://github.com/GriffonSpade/ServUO/tree/OppositionTribe-1.0)
Opposition Tribes V1.02 replaces OppositionGroup's systemic functionality.
Uses a simple TribeType switch and target's TribeType check.

BaseCreature:
-Changed Replaced functioning OppositionGroup with Opposition Tribes in Core.TOL check:
--Added enum TribeType
---Added GrayGoblin, GreenGoblin TribeTypes
--Added public virtual bool Tribe
--Added public virtual bool IsTribeEnemy()
--Added IsTribeEnemy() check in IsFriend() and IsEnemy()
-Deprecated OppositionGroup usage in !Core.TOL check in IsFriend() and IsEnemy()

Other:
-Added Tribe to Goblins 
-Removed Erroneous OppositionGroup from Goblins
-Fixed Savage Riders are always male
-Fixed Savage Shamans are always female
-Fixed Goblin names uncapitalized and articled, corpse name articles fixed

### DIFF
--- a/Scripts/Mobiles/Normal/BaseCreature.cs
+++ b/Scripts/Mobiles/Normal/BaseCreature.cs
@@ -1009,7 +1009,7 @@ namespace Server.Mobiles
         // Tribe Opposition stuff
         public virtual TribeType Tribe{ get{ return TribeType.None ; } } // What opposition list am I in?
 
-        private bool IsTribeEnemy(Mobile m)
+        public virtual bool IsTribeEnemy(Mobile m)
         {
             // Target must be BaseCreature
             if (!(m is BaseCreature))
@@ -1063,9 +1063,21 @@ namespace Server.Mobiles
 
 		public virtual bool IsFriend(Mobile m)
 		{
-			if (Tribe != TribeType.None && IsTribeEnemy(m))
+			if (Core.TOL)
 			{
-				return false;
+				if (Tribe != TribeType.None && IsTribeEnemy(m))
+				{
+					return false;
+				}
+			}
+			else
+			{
+				OppositionGroup g = OppositionGroup;
+
+				if (g != null && g.IsEnemy(this, m))
+				{
+					return false;
+				}
 			}
 
 			if (!(m is BaseCreature))
@@ -1143,9 +1155,21 @@ namespace Server.Mobiles
 				return a.IsEnemy(m);
 			}
 
-			if (Tribe != TribeType.None && IsTribeEnemy(m))
+			if (Core.TOL)
 			{
-				return true;
+				if (Tribe != TribeType.None && IsTribeEnemy(m))
+				{
+					return true;
+				}
+			}
+			else
+			{
+				OppositionGroup g = OppositionGroup;
+
+				if (g != null && g.IsEnemy(this, m))
+				{
+					return true;
+				}
 			}
 
 			if (m is BaseGuard)
--- a/Scripts/Mobiles/Normal/BaseCreature.cs
+++ b/Scripts/Mobiles/Normal/BaseCreature.cs
@@ -1021,14 +1021,14 @@ namespace Server.Mobiles
 
             switch(Tribe)
             {
-                case TribeType.Terathan: return (c.TribeType == TribeType.Ophidian);
-                case TribeType.Ophidian: return (c.TribeType == TribeType.Terathan);
-                case TribeType.Savage: return (c.TribeType == TribeType.Orc);
-                case TribeType.Orc: return (c.TribeType == TribeType.Savage);
-                case TribeType.Fey: return (c.TribeType == TribeType.Undead);
-                case TribeType.Undead: return (c.TribeType == TribeType.Fey);
-                case TribeType.GrayGoblin: return (c.TribeType == TribeType.GreenGoblin);
-                case TribeType.GreenGoblin: return (c.TribeType == TribeType.GrayGoblin);
+                case TribeType.Terathan: return (c.Tribe == TribeType.Ophidian);
+                case TribeType.Ophidian: return (c.Tribe == TribeType.Terathan);
+                case TribeType.Savage: return (c.Tribe == TribeType.Orc);
+                case TribeType.Orc: return (c.Tribe == TribeType.Savage);
+                case TribeType.Fey: return (c.Tribe == TribeType.Undead);
+                case TribeType.Undead: return (c.Tribe == TribeType.Fey);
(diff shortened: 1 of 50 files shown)


kevin-10:
Why not just keep the name as OppositionGroup, or rename it to OppositionType?  I don't want folks to get it confused with Eodon Tribes, which have their own set of opposition rules.  

The tuple and dictionary are somewhat redundant upon themselves.  You can use a simple switch statement and remove the need for the tuple and dictionary.  Here would be an example in IsTribeEnemy function:

               ```
private bool IsTribeEnemy(Mobile m) // I would prefer to keep it as IsOpposition
		{
			BaseCreature bc = m as BaseCreature;
			
			// this only applies BaseCreature vs. BaseCreature
			if(bc == null)
				return false;
			
			TribeType type = bc.Tribe;
			
			switch(Tribe)
			{
				case TribeType.None: return false; // this shouldnt happen as its already handled
				case TribeType.Terathan: return type == TribeType.Ophidian;
				case TribeType.Ophidian: return type == TribeType.Terathan;
				case TribeType.Savage: return type == TribeType.Orc;
				case TribeType.Orc: return type == TribeType.Savage;
				case TribeType.Fey: return type == TribeType.Undead;
				case TribeType.Undead: return type == TribeType.Fey;
				case TribeType.GrayGoblin: return type == TribeType.GreenGoblin
				case TribeType.GreenGoblin: return type == TribeType.GrayGoblin;
			}
		}
```

**Comment by kevin-10** [comment]
Also, instead of editing tons of files, you can alleviate some of this.  Undead and Fey opposition groups actually belong to their own slayer groups, so using the superslayer lists already described can save some time.  In BaseCreature.cs, you can have the virtual property check for one of these two types of creatures, and if not, will return the default of none.  Of course, overriding this for hte other opposition groups will remove this functionality.

		public virtual TribeType Tribe 
		{
			get
			{
				if(Array.IndexOf(SlayerGroup.Groups[1].Super.Types, GetType()) > -1)
					return TibeType.Undead;
				
				if(Array.IndexOf(SlayerGroup.Groups[6].Super.Types, GetType()) > -1)
					return TibeType.Fey;
				
				return TribeType.None;
			}
		}

**Comment by GriffonSpade** [comment]
The point of the dictionary is to frontload most of the work so that it needn't be done every time.

Equivocating slayer groups and opposition is absolutely untenable. Not only are there not slayer groups for each of them, there can be additions and exceptions.

I've been able to find ZERO information about eodon tribe interactions. Do they all attack each other? If so, that would actually make the dictionary even better for that. It would be quite simple to add them all.

**Comment by kevin-10** [comment]
I believe its the barrab tribe and myrmidex that attack the other tribes, and vice versa.  They will attack players who are of the opposite tribe vs. britannian allegience.  All this is handled in their IsEnemy function so it wouldnt be feesable to add it to the opposition system.

**Comment by kevin-10** [comment]
Fey and undead are the only groups that would mirror the slayer lists.  But yea, using the switch statement requires the same realtime checks, without having to front load any data.  Basically all it's doing is comparing one enum element to the other.

**Comment by GriffonSpade** [comment]
Fey and Undead are one of the main reasons why we can't mirror the slayer lists. 

There are 'dark' fey that pop up with undead creatures. 

And it may not be appropriate to include all undead slayers as undead opposition, as they may be uninvolved or isolated.
It may also be appropriate for non-undead to be in the undead opposition, like necromancer types and shadow-creatures.

File: Scripts/Mobiles/Normal/BaseCreature.cs [hunk 1]
```diff
@@ -993,6 +1006,33 @@ public virtual void BeginFlee(TimeSpan maxDuration)
         public virtual OppositionGroup OppositionGroup { get { return null; } }
         public virtual bool IsMilitiaFighter { get { return false; } }
 
+        // Tribe Opposition stuff
+        public virtual TribeType Tribe{ get{ return TribeType.None ; } } // What opposition list am I in?
+
+        private bool IsTribeEnemy(Mobile m)
```
Vorspire: Why is this private?

GriffonSpade:
@Vorspire The issue with OppositionGroup is that it's not an enum, it's an ArrayList containing both oppositions, each of which is a list containing all the members of that opposition. 
It first checks which opposition the creature is on by scanning every member of both opposition lists until it finds it and assigns its group. Then it checks which opposition the target is on by scanning every member of both the first creature's opposition lists until it finds it and assigns its group. (including invalid) It then returns false if either are invalid or they are equal.
I think it may be the most obtuse and convoluted way to be done.

My system by comparison assigns each opposition an enum tribe.
It first checks if the target is a player and returns false if it is. Then it enters a switch that picks the creature's current tribe, and returns false if the target is not in the creature's opposition's enemy opposition.
It requires no lists of opposition members and can easily be changed to support multiple enemy oppositions for each given opposition. (Including adding a dictionary)

As for IsTribeEnemy() being private, my only defense is that it was 3-4 AM when I wrote that.

Otherwise, I've taken your statement to heart to create a Core.TOL check to use IsEnemyTribe() instead of OppositionGroup.IsEnemy(), and contain the latter in the else.

I did not readd the SavagesAndOrcs oppositiongroup back to the goblins because it's copy-pasta: Goblins have their own thing against each other, and no such oppositiongroup exists as of right now.

Vorspire:
I definitely agree with your rewrite more than the old way of doing it, if both systems are functionally the same, I don't see why we can't phase out the OppositionGroups over time now. It's just a bit much to to throw out a (quite large) breaking change without warning hehe.

kevin-10:
I was going to suggest simply rename your enum to OppositionGroup, and remove all the existing OppositionGroup stuff for compatibility, however IsEnemy references an instance of OppositionGroup so that wont work!

GriffonSpade:
Yeah, it would also be a hard error for any custom OppositionGroup stuff that way. And let's not forget the weirdness that both the type and value are both called OppositionGroup, lol. Messing with OppositionGroup is a real head-banger. Plus the new one isn't really an Opposition 'group', since it's a single opposition instead of two :P. Best to leave OppositionGroup as-is, and transition to something new.

PR #1926, opened 2017-04-04
Fixes to Cleansing winds mystic spell

ppardalj:
This pull requests add some fixes to *Cleansing Winds* mystic spell, as per OSI tests. These fixes include:
- Triggers `OnBeneficialAction` for all targets. This means if heal affects a criminal/murderer player in the party, the caster will be flagged as criminal.
- Reduced area of effect from 3 tiles to 2 tiles.
- Fixed base healing formula as per OSI: (mystic+focus)/4
- Reduced maximum amount of possible additional targets to 3.
- Added healing reduction per additional target.
- Fixed healing reduction factor due to poison: now it's 15% per level of poison.
- Added a graphical effect to every affected player.

fyi @GriffonSpade @kevin-10


kevin-10:
It does not compile.  :(

ppardalj:
@kevin-10 it does compile for me. Can you please post your errors?

ppardalj:
@kevin-10  that is C# 6.0 syntax. As per https://www.servuo.com/threads/servuo-is-upgrading-to-net-4-6-1.5840/ this project is now targeting .NET Framework 4.6.1 which support C# 6.0 features, so this code should be valid. Please upgrade your environment and check again.

Tasanar:
The problem is ServUO has not officially done this yet. It is one of the branches but it is not merged. 

@dmurphy22

GriffonSpade:
public override SpellCircle Circle => SpellCircle.Sixth;

is equal to:

public override SpellCircle Circle { get { return SpellCircle.Sixth; } }

Is that correct? Using, ah, extension body instead of braces?

kevin-10:
Yes, that is correct.  Maybe @dmurphy22 can chime in, but I don't think we're at the point where we're forcing others to upgrade.

ppardalj:
@kevin-10 @GriffonSpade the branch that is not merged yet is support for code compilation using the [Roslyn](https://github.com/dotnet/roslyn) compiler, but nothing prevents you from using 4.6.1 with the legacy compiler.

In any case let's wait to hear @dmurphy22 feedback, but I honestly think this should not be a blocker.

ppardalj:
@kevin-10 @Tasanar I have removed the usage of this C# 6.0 syntax. Could you please give another try to this pull request?

PR #1927, opened 2017-04-05
Generalize IsBlackSolen/IsRedSolen

GriffonSpade:
EDITING: DO NOT MERGE.

The purpose of this pull is to generalize black and red solen groups so that when they are checked in BaseAI, it does not check a type list of them.

-Added virtual bool IsBlackSolen (default false)
--Add override true in Black Solen creatures
-Added virtual bool IsRedSolen (default false)
--Add override true in Red Solen creatures
-Changed Cleaned up EndPickTarget().if (order == OrderType.Attack) checks
-Changed Solen checks to check IsBlackSolen and IsRedSolen bools instead of a list of types.

### DIFF
--- a/Scripts/Mobiles/AI/BaseAI.cs
+++ b/Scripts/Mobiles/AI/BaseAI.cs
@@ -331,27 +331,28 @@ namespace Server.Mobiles
 
 			if (order == OrderType.Attack)
 			{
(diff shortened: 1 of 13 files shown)


GriffonSpade:
This does sound superior. The interfaces can be added to SolenHelper.cs, where they would be with the other properties checked.

GriffonSpade:
Closing this, will just open a new PR.

PR #1931, opened 2017-04-05
Mobile.Murderer property

ppardalj:
This pull request introduces the property `Mobile.Murderer` which returns `true` when `Kills >= 5`. This has been introduced to remove all the duplicated checks like that across the whole codebase. In a later pull request, this value may become configurable.

### DIFF
--- a/Scripts/Gumps/AdminGump.cs
+++ b/Scripts/Gumps/AdminGump.cs
@@ -159,7 +159,7 @@ namespace Server.Gumps
                 case AccessLevel.Player:
                 default:
                     {
-                        if (m.Kills >= 5)
+                        if (m.Murderer)
                             return 0x21;
                         else if (m.Criminal)
                             return 0x3B1;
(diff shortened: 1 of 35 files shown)


GriffonSpade:
In BaseCreature, it should probably check AlwaysMurderer. 

ie:
public override bool Murderer { get { return AlwaysMurderer || m_Kills >= 5; } }

Are there any conflicts that would arise from doing so?

ppardalj:
@GriffonSpade you're probably right about the BaseCreature check but I didn't want to include it due to the possible collateral effects, so I decided to keep it simple this time. But will definitely look into it in the future, although not for this PR.

kevin-10:
It's fine either way.  The only time AlwaysMurder is used is for notoriety reasons.  Past practice, is if you want a creature to appear red, you override this property instead of marking its kills to 5.  Making it red by kills can, in theory, have the opposite effect if the creature lives long enough and burns off its murder counts.

PR #1932, opened 2017-04-05
Saves and Logs directories are now configurable

ppardalj:
This pull request adds the ability to configure the Logs/, Saves/ and Backups/ directory location, so that shard admins can store their world saves, backups and logs in a location different to ServUO root directory, in order to achieve things such as for example:
- A faster disk in order to improve save times
- A dedicated, more capacity disk, for backups
- Additional log streams.


GriffonSpade:
This also sounds like it would be useful for testing: I don't need a new save directory for each non-serialized variation, and it saves me from manually having to migrate my saves around. Still have to migrate configurations, but that's unavoidable.

ppardalj:
@GriffonSpade Backups are stored always in <basedir>/Backups directory. It could be a possibility also to make it configurable, actually seems that the needs it's bigger since backups quickly consume up the disk space :P I will take a look on that. Thanks for the idea!

ppardalj:
@GriffonSpade said and done! just pushed a commit to make Backups directory also configurable. By default it's {BaseDirectory}/Backups, but you can edit it via config to point it to the directory you want.

Xeroxxx:
Why a new config file for every path? Just add it to general.cfg or put it all in one.

ppardalj:
@Xeroxxx each config is intented to contain not only the directory, but any other config parameter related to that feature, but those are out of the scope of this pull request.

Xeroxxx:
Okay but I don't think its more user-friendly to add more configs. I think the future should be consolidate as much configs as possible.

However I see a lot of changes in your code that are note necessary or documented. Lots of stings changed to var's etc. Can you explain whats the benefit?

Cheers

ppardalj:
@Xeroxxx I understand your concern, it's true that it may complicate things at this moment, but I also think that there's a point on keeping each feature configuration separated, I would ask for more input here. How would you name this config file then?

Regarding the rest of changes, they are purely cosmetic, now that the project is using the latest versions of C# which contains a lot of syntactic sugar like the type inference, which enhances the readability of the code, I just took the liberty to use it in the code I changed. It's true that it's no strictly related with the PR, but they are pretty straightforward changes.

PR #1948, opened 2017-04-07
Filters .editorconfig to C# source files

jdphenix:
This should resolve future issues with the existing .editorconfig interacting with editors to change whitespace in unintended source files. 

This change modifies the .editorconfig settings to only apply to C# source files.


GriffonSpade:
Ah, great. That should help.

PR #1952, opened 2017-04-08
Nether Bolt Stack Damage Fix

kevin-10:
- Updated spell effects on HealingStone and NetherBolt
- Nether Bolt will no longer stack damage, like magic arrow. Confirmed on EA.
- TODO: Rest of mystic Spells and pre-cast effects need to be moved to spell timer. this is for chiv, mystic and skill mastery Spells


kevin-10:
Yes, the mechanics should be.  As far as the effects, I don't think they were ever correct on ServUO.

PR #1962, opened 2017-04-11
* Add new tameable creatures

Argalep:


GriffonSpade:
Why do these warrant checks to see if they're not controlled when spawned? Is there a way for players to somehow spawn these creatures already controlled? Otherwise, hiding these pack items in such checks is meaningless.

Also, 'this.' is extraneous.

Overall, looks good :)

Scripts/Mobiles/Normal/DreadSpider.cs
```diff
@@ -14,64 +14,59 @@ public DreadSpider()
...
+            this.SetSkill(SkillName.Necromancy, 20.0);
+            this.SetSkill(SkillName.SpiritSpeak, 20.0);
 
             this.Fame = 5000;
             this.Karma = -5000;
 
             this.VirtualArmor = 36;
 
-            this.PackItem(new SpidersSilk(8));
+            if (!this.Controlled)
```
kevin-10: Creatures will never be controlled once they are constructed.  You can remove these checks.  Also, for the shadow wyrm and Pheonix, they were made tamable on EA with the new pet training publish, correct?  So, would the new pet training that is missing be an issue?

Argalep: I just added the creatures. How could it be a problem

PR #1967, opened 2017-04-12
Consecrate Weapon and Enemy of One damage modifier changes

bsenyuva:
Consecrate Weapon: Based on chiv skill will give damage bonus 1-15 when skill value >= 92 as OSI Test Center
Enemy of One: Based on chiv skill will give damage modifier bonus up to 64. Karma is not considered yet. Formula taken from http://stratics.com/threads/chivalry-changes.245029/#post-1977380

### DIFF
--- a/Scripts/Items/Equipment/Weapons/BaseWeapon.cs
+++ b/Scripts/Items/Equipment/Weapons/BaseWeapon.cs
@@ -2230,11 +2230,15 @@ namespace Server.Items
 
             if (m_Consecrated)
             {
-                if (Utility.RandomDouble() <= ConsecrateProcChance / 100)
-                {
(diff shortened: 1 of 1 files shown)


bsenyuva:
I've created a variable to cache value in a literal instead of function call from object for 3 times.

bsenyuva:
I've removed commented out code.

kevin-10:
Cool, thanks.  Revert the chiv skill variable back as well.

bsenyuva:
Done. It will be use more CPU for every weapon hit on server.

kevin-10:
Actually, no it won't.  I'd say the number of weapon hits that involve these paladin spells are less than 1%.  Far less.

PR #1970, opened 2017-04-13
Updates to Enemy of one formula

ppardalj:
This pull request:
- Updates Enemy of one spell damage bonus formula to OSI.
- Apply only new damage bonus calculation only if post-SA era, otherwise use classic bonus (always 50).
- Refactored spell implementation to remove dependency with PlayerMobile class. Now most spell mechanics are implemented in EnemyOfOne.cs

### DIFF
--- a/Scripts/Items/Equipment/Weapons/BaseWeapon.cs
+++ b/Scripts/Items/Equipment/Weapons/BaseWeapon.cs
@@ -94,7 +94,7 @@ namespace Server.Items
         }
 
 		/* Weapon internals work differently now (Mar 13 2003)
-        * 
+        *
         * The attributes defined below default to -1.
         * If the value is -1, the corresponding virtual 'Aos/Old' property is used.
         * If not, the attribute value itself is used. Here's the list:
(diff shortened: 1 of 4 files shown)


kevin-10:
There is a small effect issue when you toggle EOO off.  Also, when it's first cast, the damage scalar isn't showing the proper damage increase.  I would assume this should show when you cast it, correct?

ppardalj:
@kevin-10  what is the issue? For me looks similar to OSI, both "activate" and "deactivate" effects are played together.

Regarding damage scalar, it's updated when you first hit a target. OSI also behaves this way.

kevin-10:
The pre-cast effects last too late night when the cast delay is set to TimeSpan.Zero for some reason.  In my testing on EA, when EOO is active, the delay is typically 1/2 of the normal cast time, not removed altogether. So, cutting the delay in half should fix the cosmetic issue.  Also, when a EOO target is determined, the cliloc on the buff icon changes to take into account what type of creature it is.  For example, a mongbat shows as mongbats so this is going to require some string hacking.

kevin-10:
Late night = long

PR #1975, opened 2017-04-14
Tithing point gump fix, per EA.

kevin-10:


kevin-10:
@ppardalj, please see my work-around for the buff icon for Enemy of One.  Thanks!

PR #1979, opened 2017-04-15
Familiar AI Fix/Revamp v4

GriffonSpade:
[Associated Thread](https://www.servuo.com/threads/familiar-ai-fix-revamp.6961/)
[Associated Branch](https://github.com/GriffonSpade/ServUO/tree/Familiar-AI-Fix/Revamp)

This version 4 should have full OSI fidelity.

-Familiars with AttacksMastersTarget ONLY attack creatures within RangeHome (default 10) of master.

v2:
-Fixed a a bug causing familiar to not attack anything not attacking itself
-Various incidental Trims

v3:

v4:
-Added Familiars without AttacksMastersTarget now have melee-range combat as per OSI. (Both require familiar to be within 1 range of master, and target can be within 1 range of either the familiar or the master)

### DIFF
--- a/Scripts/Mobiles/AI/BaseAI.cs
+++ b/Scripts/Mobiles/AI/BaseAI.cs
@@ -1179,6 +1179,15 @@ namespace Server.Mobiles
 				case OrderType.Transfer:
 					return DoOrderTransfer();
 
+				case OrderType.Aggro:
+					return CheckFamiliar();
+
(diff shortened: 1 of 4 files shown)


Tasanar:
What I like about this is if it really does fix the issue with some familiars. A good example is the shadow wisp summons. It will lose its master all the time, usually on log out, and just spasm out in its current spot.

PR #1980, opened 2017-04-16
Removed IUsesRemaining from axes. 

kevin-10:
Confirmed on EA that axes DO NOT BREAK from lumberjacking


jdphenix:
Do not axes break from use on siege perilous?

kevin-10:
Yes, according to UOGuide, but not on regular production shards.  I used 500+ swipes and no breakage.

GriffonSpade:
...Do we have a configuration setting for Siege Perilous style? If not, perhaps a Milestone Task?

kevin-10:
That would be a monumental task but I'd be down.

PR #1984, opened 2017-04-17
Bug fixes/Updates

kevin-10:
- Mischief Maker is now undead Slayer
- Fixed Cliloc on Katalkotl's Ring
- Added Skill Pick gump to Katalkotl's Ring
- Removed Sleep Property, timer, etc from Mobile.cs as it is not used
- Updated consecrate weapon, now used ConsecratedWeaponContext, like EnemyOfOne, all data is accessible through the context. The last update left out some of the newer features of the spell, and had some unused properties.

Brought several items/systems in line with Publish 71:
http://www.uoguide.com/Publish_71
- Consecrate Weapon
- Bandage changes
- Enchanted Apple Changes
- Poison Changes
- Orange Petal Changes
- Lightning Strike


Scripts/Spells/Bushido/LightningStrike.cs
```diff
@@ -13,7 +13,7 @@ public override int BaseMana
         {
             get
             {
-                return 5;
+                return 10;
```
ppardalj: `return Core.SA ? 10 : 5`

kevin-10:
Yes, there are a few spots that will need that.  Thanks!

PR #1986, opened 2017-04-17
Update LargeBulkEntry.cs

Argalep:
Crash Fix.


kevin-10:
Wow, how did that get by!  Thanks.

PR #2049, opened 2017-05-03
Added mouse over properties to PlayerMobile, per EA.

kevin-10:


Argalep:
Only valid for test center

kevin-10:
Oh I see, thanks!  I'll add a test center check.

PR #2051, opened 2017-05-04
BaseAI.cs: moved detect hidden from AITimer into AI proper

GriffonSpade:
-Added AfterThink() method, containing detect hidden check, called in OnTick() where the detect hidden check was formerly. Code style fix, no functional changes for unmodified rules, but allows the check to be overridden in inherited AI.


GriffonSpade:
Yeah. I'm working on another version of this. It's not actually even using the skill itself, just its values. It will be called FindPlayer to distinguish it from actual uses of the detect hidden skill. (It ONLY finds hidden players) It will also have a boolean to even check it, which will be in OnThink().

PR #2061, opened 2017-05-07
MageAI: Fleeing Bugfix

GriffonSpade:
Overview:
Alright, so, apparently this problem is larger than I thought. It's pretty much the same 3-4 issues copy-pasted into almost every single AI. With the exception of AnimalAI and BaseAI, all of them changed received basically the same changes, with the occasional different fleeing condition or special fleeing ability. Basically, using FocusMob the way it was done here doesn't really work for the most part. It is constantly reset and can only be acquired once every ten seconds normally, meaning it either breaks the AI's ability to flee or just uses combatant anyway.

MageAI, ArcherAI, MeleeAI, SpellbinderAI, OmniAICore, ThiefAI, OrcScoutAI, AnimalAI, BaseAI

-Fixed Removed redundant 'this.' from all individual AI files
-Fixed removed redundant Debug checks for DebugSay from all individual AI files
-Changed IDamageable m_Mobile.Combatant's local variable to 'c' in DoActionCombat
-Fixed and simplified flee check algorithm on most individual AIs
--Many had inverted algorithm before, making it LESS likely for mobile to flee if combatant had more health. It was, in fact, IMPOSSIBLE for the creature to flee if its combatant had more than 90 hits more than it. Others were checking !IsParagon instead of CanFlee.
-Added When creatures are ready to reenter combat, they check if they have an active combatant, and if they do, they reengage it instead of going to guard mode
-Changed When most creatures are fleeing, they instead check new BaseAI flee code that is not based on checking FocusMob (which doesn't work except once every 50 or so AI ticks usually) but instead combatant
-Changed When creatures with abilities they use while fleeing, they use them in the same else check as base.DoActionFlee()
-Removed redundant settings when setting ActionType, which are already done in OnActionChanged
-Removed setting FocusMob for base.DoActionFlee's usage. It uses combatant now.

BaseAI.cs:
-DoActionFlee()
-Changed adapted with fleeing based on that made for MageAI, and is now called by most AI
--Fixed Now checks acquirefocusmob OR has a combatant within double perception range to run away from when fleeing.
---Required successful check of acquirefocusmob every AI tick before, which can only happen every 10 seconds normally
--Changed Now only attempts to acquire Aggressor fightmode targets: When a creature is fleeing it only wants to acquire things that will attack it, not things it wants to attack
---Fixed now sets combatant to focus mob if it found one, and then flees from combatant
---Added Added some bad target abort checks that remove combatant
-Added range check to reengage combatant after recovering rather than guarding

AnimalAI.cs: Completely removed DoActionFlee override, uses base completely.

ThiefAI.cs: Completely removed DoActionFlee override, uses base completely. (It was set to exit flee action when at high health, breaking its steal-and-flee ability) (Its steal ability still doesn't work very well, but this is primararily for fleeing)

ArcherAI.cs, OrcScout.cs:
-Added reference System
-DoActionCombat()
--Fixed nested, but commented ammo flee check inside the main able to flee check (They keep shooting anyway)
-Added DoActionFlee()
--Based on MageAI.cs with BaseAI.cs's movement method
---Commented out: Continues to flee if creature lacks ammo (They can shoot without ammo)

-OrcScout.cs: Added CanStealth, Tribe 		

-Stealth.cs: Removed playermoble requirement
--IsStealthing has long since been moved from playermobile to mobile

-OrcChopper.cs: removed 'this.', lowercased name, added Tribe 		

-OppositionGroup.cs: Added OrcChopper, OrcScout 		

Some post-process notes:
Creatures still keep using their bows (successfully!) even without any ammo, but that's fairly well known.
Why do animals flee at just under 10% health while other AI flee starting at just under 20% health?

### DIFF
--- a/Scripts/Mobiles/AI/MageAI.cs
+++ b/Scripts/Mobiles/AI/MageAI.cs
@@ -75,10 +75,10 @@ namespace Server.Mobiles
 
         public override bool Think()
         {
-            if (this.m_Mobile.Deleted)
+            if (m_Mobile.Deleted)
                 return false;
 
-            if (this.ProcessTarget())
+            if (ProcessTarget())
                 return true;
             else
                 return base.Think();
@@ -91,36 +91,36 @@ namespace Server.Mobiles
 
         public virtual double ScaleBySkill(double v, SkillName skill)
         {
-            return v * this.m_Mobile.Skills[skill].Value / 100;
+            return v * m_Mobile.Skills[skill].Value / 100;
         }
 
         public override bool DoActionWander()
         {
-            if (this.AcquireFocusMob(this.m_Mobile.RangePerception, this.m_Mobile.FightMode, false, false, true))
+            if (AcquireFocusMob(m_Mobile.RangePerception, m_Mobile.FightMode, false, false, true))
             {
-                this.m_Mobile.DebugSay("I am going to attack {0}", this.m_Mobile.FocusMob.Name);
+                m_Mobile.DebugSay("I am going to attack {0}", m_Mobile.FocusMob.Name);
 
-                this.m_Mobile.Combatant = this.m_Mobile.FocusMob;
-                this.Action = ActionType.Combat;
-                this.m_NextCastTime = DateTime.UtcNow;
+                m_Mobile.Combatant = m_Mobile.FocusMob;
+                Action = ActionType.Combat;
+                m_NextCastTime = DateTime.UtcNow;
             }
-            else if (this.SmartAI && this.m_Mobile.Mana < this.m_Mobile.ManaMax && !this.m_Mobile.Meditating)
+            else if (SmartAI && m_Mobile.Mana < m_Mobile.ManaMax && !m_Mobile.Meditating)
             {
-                this.m_Mobile.DebugSay("I am going to meditate");
+                m_Mobile.DebugSay("I am going to meditate");
 
-                this.m_Mobile.UseSkill(SkillName.Meditation);
+                m_Mobile.UseSkill(SkillName.Meditation);
             }
             else
             {
-                this.m_Mobile.DebugSay("I am wandering");
+                m_Mobile.DebugSay("I am wandering");
 
-                this.m_Mobile.Warmode = false;
+                m_Mobile.Warmode = false;
 
                 base.DoActionWander();
 
                 if (Utility.RandomDouble() < 0.05)
                 {
-                    Spell spell = this.CheckCastHealingSpell();
+                    Spell spell = CheckCastHealingSpell();
 
                     if (spell != null)
                         spell.Cast();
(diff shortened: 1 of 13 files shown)


GriffonSpade:
I found similar issues in other AI. I'm going to add any related BaseAI, ArcherAI, and MeleeAI fixes to this pull.

Scripts/Mobiles/AI/MageAI.cs
```diff
@@ -790,53 +775,77 @@ public override bool DoActionGuard()
...
+            if ((m_Mobile.Mana > 20 || m_Mobile.Mana == m_Mobile.ManaMax) && m_Mobile.Hits > (m_Mobile.HitsMax / 2))
             {
-                this.m_Mobile.DebugSay("I am stronger now, my guard is up");
-                this.Action = ActionType.Guard;
+                // If I have a target, go back and fight them
+                if (c != null)
+                {
+                    m_Mobile.DebugSay("I am stronger now, reengaging {0}", c.Name);
+                    Action = ActionType.Combat;
+                }
```
kevin-10: Lets do a range check, so if combatant is not null, however out of range, it can go into guard.  Most range checks in AI are perception x2 i believe.

GriffonSpade: Roger that. I've expanded it a lot, but I'll go back and fix up that check on each file with a range check. I just didn't think about it because it's checked at the previous tick while still fleeing.

GriffonSpade:
Alright. Range check implemented to reengage rather than guard. I did not expect a 20 minute single-method fix to turn into a 6 hour marathon of fixing over half a dozen different AI with pretty much the same problems with slightly different details.

GriffonSpade:
The slightly different details. Mana checks, special abilities, some not even using anything else than what's in BaseAI right now. What got shoved into BaseAI.DoActionFlee() actually wasn't /completely/ universal before, but it was close enough that it was creditable to the fact that they were made ad hoc by different people years apart, and just needed rounding out and polishing.

kevin-10:
Orc Scouts should not stay hidden when they shoot.  They do not on EA.

GriffonSpade:
Yeah, I don't even understand how they're staying hidden. Isn't attacking a revealing action? How can a creature be forced to not attack without removing the combatant?

GriffonSpade:
Oh, might there be something with Warmode? I'll check it out.

GriffonSpade:
Oh, I might have been mistaken, and my highlighted attackable target being the same as hiding threw me.

GriffonSpade:
Alright, I think that makes everything flee at least decently. Orc Scout still keeps shooting while fleeing and breaking its stealth, but that's likely unavoidable, and not necessarily wrong. Because orcs are stupid. And I doubt anyone wants to track down mobiles that stealth away when almost dead.

Everything else looking solid.

PR #2062, opened 2017-05-07
Bug fixes

kevin-10:
- PurgeMagicSpell curse now doubles mana cost until the target does damage
- Curse/Weaken/Clumsy/Feeblemind now takes priority depending on which has the highest stat malus
- Removed DisableSkillCheck static property when adding curses and buffs. it is replaced by a simple argument
- Co-Owners can now place addons in a house
- Co-Owners who place addons can now remove those addons, as well as the Owner. Addons was replaced with a dictionary to keep track of who placed it
- replaced arraylists in BaseHouse.cs to generic list.
- fixed T-Map pin position on server restart

Peerless minor arty drop fix
- minor arties now drop on peerless corpse
- appropriate arties will no longer be elf restricted
- fixed hues of various minor arties


Tasanar:
Why the removal of the Medusa altar?

Argalep:
MedusaAltar is not used. Correct use MedusaPlatform.

PR #2067, opened 2017-05-09
Fixs

Argalep:
* CornStok Fix
* Blue Corn hue fix
* Exploring Quest fix. EA
* IgnisFatalis fix

**Note**: Delete the Sorcerers Reward Chest and add it again.

### DIFF
--- a/Scripts/Items/Consumables/EodonPotions.cs
+++ b/Scripts/Items/Consumables/EodonPotions.cs
@@ -780,7 +780,7 @@ namespace Server.Items
         [Constructable]
         public BlueCorn(int amount)
         {
-            Hue = 1282;
+            Hue = 1284;
             Amount = amount;
         }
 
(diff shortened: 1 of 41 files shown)


Scripts/Items/Consumables/EodonPotions.cs
```diff
@@ -802,30 +802,61 @@ public override void Deserialize(GenericReader reader)
         }
     }
 
-    public class BlueCornStock : Item
```
kevin-10: Are these on spawners or do they have an internal spawner?  If so, the spawners need to be changed (in the spawner file) as well as changing the name on any existing spawners.  See GiantSerpent.cs for how to do this.  Or, you can just keep the name as BlueCornStock to avoid this.

Argalep: I did not see a place where it was used

GriffonSpade: @Line 807 that you added: public override int LabelNumber { get { return 1035639; } } // corn stalk

Also, backwards compatibility is needed not just for code for spawners, etc. but also for shards that have existing blue corn items spawned, otherwise they will be forced to delete them. (Unless this is an all new item, or something that respawns on the ground and cannot be picked up by players)

Edit: Okay, I've looked at this in game, and it does not appear to be something players can pick up, so you can probably change the name without much issue(The admins of shards that have them spawned will still be prompted to remove them from game, though), however you need to change it to CornStalk rather than CornStock. No point in changing it from one erroneous name to another.

PR #2118, opened 2017-05-20
fixed gargish reward robes where they wouldnt be created when selected

kevin-10:


kevin-10:
No, drops should not be limited region.  You can set the NoKillRewards if you don't want an arty drop.

Scripts/Mobiles/Normal/DemonKnight.cs
```diff
@@ -171,9 +171,10 @@ public static void HandleKill(Mobile victim, Mobile killer)
 
             if (!Core.AOS)
                 return;
-            //CHeck if players are in the region with the bosses
-            if ( pm == null || bc == null || !CheckLocation(bc) || !CheckLocation(pm))
+
+            if ( pm == null || bc == null || bc.NoKillAwards/*|| !CheckLocation(bc) || !CheckLocation(pm)*/)
```
kevin-10: No, drops should not be limited region. You can set the NoKillRewards if you don't want an arty drop.

PR #2167, opened 2017-06-01
Jumu's Sacred Hide props updated

bsenyuva:
Jumu's Sacred Hide props updated according to http://www.uoguide.com/Jumu%E2%80%99s_Sacred_Hide

Poison Eater 15% added
Fire resist 5% fixed on Gargoyle version


kevin-10:
This commit gives compile errors.

PR #2176, opened 2017-06-02
Update BaseWeapon.cs

Argalep:
Imbuing BaseWeapon GetMaxWeight fix.


kevin-10:
Actually, it needs to be IResource.  IQuality derives from IResource.  IResource is where the bods get the resource from.

PR #2185, opened 2017-06-05
* add Lasher

Argalep:
* It is instantly bonded when claimed

### DIFF
--- a/Scripts/Mobiles/Normal/BaseCreature.cs
+++ b/Scripts/Mobiles/Normal/BaseCreature.cs
@@ -2811,7 +2811,12 @@ namespace Server.Mobiles
 
                             if (master != null && master == from) //So friends can't start the bonding process
                             {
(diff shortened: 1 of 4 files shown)


Scripts/Mobiles/Normal/BaseCreature.cs
```diff
@@ -2811,7 +2811,12 @@ public virtual bool CheckFeed(Mobile from, Item dropped)
 
                             if (master != null && master == from) //So friends can't start the bonding process
                             {
-                                if (m_dMinTameSkill <= 29.1 || master.Skills[SkillName.AnimalTaming].Base >= m_dMinTameSkill ||
+                                if (this is Lasher || this is Windrunner)
+                                {
+                                    IsBonded = true;
```
kevin-10: We can set IsBonded in ConfirmMountStatuetteGump.cs to keep BaseCreature clean.

PR #2198, opened 2017-06-10
fixs.

Argalep:
* Cleanup Reward MonsterStatuette  fix.
* Veteran Reward MonsterStatuette fix.


kevin-10:
Why the params in CollectionItem?

Scripts/Items/Decorative/MonsterStatuette.cs
```diff
@@ -184,6 +171,8 @@ public class MonsterStatuette : Item, IRewardItem
         private MonsterStatuetteType m_Type;
         private bool m_TurnedOn;
         private bool m_IsRewardItem;
+        private Timer m_NewsTimer;
```
kevin-10: So, if we're using the old Sherry the Mouse script, while are we adding Town Crier functionality to MonsterStatuette?

Argalep: I could not decide.  It seems more accurate to use the code in MonsterStatuette

kevin-10: Ther is no need for the town crier functionality in MonsterStatuette.

PR #2212, opened 2017-06-17
Bug Fixes and things

kevin-10:
- Coral the Owl now only shows vet reward if its a vet Reward
- Fixed new BOD Talismans
- fixed cosmetic issues in BOD reward gumps
- Huntsmaster system now gives points for top hunts for each Species
- Redeeming huntsmaster rewards now only yields 1 random reward. Still waiting on feedback for the rewards, and how they are chosen.

### DIFF
--- a/Scripts/Items/Artifacts/Equipment/Jewelry/AnkhPendant.cs
+++ b/Scripts/Items/Artifacts/Equipment/Jewelry/AnkhPendant.cs
@@ -31,7 +31,7 @@ namespace Server.Items
         public AnkhPendant()
             : base(0x3BB5)
         {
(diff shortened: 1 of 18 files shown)


Tasanar:
In regards to the newest changes - I am going to test now but one thing I see that is possible, will it FORCE addons to show? like a forge?

Tasanar:
Two errors are thrown while trying to start up an existing shard with this. I posted the issues in their respected posts on ServUO.

kevin-10:
Should be good now.  The one for basecreature and playermobile, you must have reverted your mobile.cs somehow.  Make sure, in Mobile.cs, AddItem method is marked virtual:

[code]public virtual void AddItem(Item item)[/code]

PR #2288, opened 2017-07-03
Toolbar.cs: Sanity check for null netstate to prevent crash

GriffonSpade:
I believe the problem is that it is trying to send the admin toolbar to a player that is not online, thus having no netstate, causing a crash.


kevin-10:
I added a null check with my last commit.  Thanks!

PR #2304, opened 2017-07-07
Toolbar.cs: Sanity check, abort sending gump to offline player

GriffonSpade:
Previous change did not fix the crash. No point in even trying to send a gump to a player that is not connected. New check aborts immediately if player is not connected.


kevin-10:
Just a question, but why does this gump get sent if the person is not online?  Maybe there is a root cause that needs to be fixed, ie if this can happen to other gumps.

GriffonSpade:
I assume it's sent to any player with admin rights. I'll look into it and see if I can find it.

PR #2306, opened 2017-07-08
Bug Fixes

kevin-10:
- Massive now places all items at 125 str Req
- Unwieldly now places items at 50 stones
- Renamed ComunityCollectionGump to CommunityCollectionGump


Tasanar:
Will this update force previously spawned items with the Massive and Unwieldy property to update? If not what would be the best way to code it in Serialization.

kevin-10:
It should auto-reflect on all items.  No deserialization changes needed.

Scripts/Services/CommunityCollections/BaseCollectionMobile.cs
```diff
@@ -367,7 +367,7 @@ public virtual void Reward(PlayerMobile player, CollectionItem reward, int hue)
 			
             reward.OnGiveReward(player, this, hue);	
 			
-            player.SendGump(new ComunityCollectionGump(player, this, this.Location));
```
Argalep: missing CommunityCollectionGump

kevin-10: Ill add the updated csproj in a few.

PR #2308, opened 2017-07-08
Toolbar crash fix: improved

GriffonSpade:
ToolbarCore.cs: Added null netstate abort
 #-Prevents crashes due to offline staff characters being killed and a gump attempting to be sent to a non-existant client.

Toolbar.cs: Reverted File, removed ad hoc sanity check and file returned to original configuration.

At Dexter's advisement, source was located very easily with GitHub's search function (much better than what I can do at home offline). Was an EventSink, specifically, that didn't abort when a staff character was killed while not connected, trying to send the gump to a non-existant client causing a crash. Now also checks if there's a null netstate to abort before sending gump.


kevin-10:
Word!

PR #2330, opened 2017-07-09
Familiar AI Fix/Revamp v5

GriffonSpade:
[Associated Thread](https://www.servuo.com/threads/familiar-ai-fix-revamp.6961/)
[Associated Branch](https://github.com/GriffonSpade/ServUO/tree/Familiar-AI-Fix/Revamp-v5)

v1:
-Familiars with AttacksMastersTarget ONLY attack creatures within RangeHome (default 10) of master.

v2:
-Fixed a bug causing familiar to not attack anything not attacking itself
-Various incidental Trims

v3:

v4:
-Added Familiars without AttacksMastersTarget now have melee-range combat as per OSI. (Both require familiar to be within 1 range of master, and target can be within 1 range of either the familiar or the master)

v5:
[s]-Fixed a bug causing erroneous lingering familiar control slot after moving to internal map and logging back in. (Which goes away after another relog) When summoner is moved to internal map, familiar is deleted, dropping any loot on the ground.[/s]
-Brought in line with current functionality, where summoned familiar remains in the world. Needs information on actual OSI behavior. Do they remain in the world as is current behavior, die and drop what they're carrying, or do they get moved to internal map with players like normal pets?

### DIFF
--- a/Scripts/Mobiles/AI/BaseAI.cs
+++ b/Scripts/Mobiles/AI/BaseAI.cs
@@ -2103,10 +2103,9 @@ namespace Server.Mobiles
             // Part 1: Delete Familiar check
             BaseCreature m = m_Mobile;
             Map map = m.Map;
-            Mobile mr = m.ControlMaster;
-            Map mrMap = mr.Map;
+            Mobile master = m.ControlMaster;
 
-            if (mr == null || mr.Deleted || mrMap == null || mrMap == Map.Internal)
+            if (master == null || master.Deleted)
             {
                 if (map != null && map != Map.Internal)
                 {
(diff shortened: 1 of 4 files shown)


GriffonSpade:
Hrm, I'm going to change that last thing to not delete the familiar, but instead leave it in world, since that's what's done right now.

kevin-10:
I just tested this for over an hour. The familiar simply takes a swing every time it's master hits via melee. This seems way too much for a simple thing as that.  All you need is a hook in your base weapon onhit, or somewhere that on hit calls to check for a familiar.

GriffonSpade:
Alright, changed the player moving to internal map behavior of familiar to match current behavior. However, it seems odd, so OSI's actual functioning is needed to be researched for the future.

This version is ready for testing and feedback.

GriffonSpade:
> I just tested this for over an hour. The familiar simply takes a swing every time it's master hits via melee. This seems way too much for a simple thing as that. All you need is a hook in your base weapon onhit, or somewhere that on hit calls to check for a familiar.

Please clarify? Familiar should be attacking current combatant, whether it's double clicked from afar, in melee range, with a spell, or auto-aggro when freshly aggressed without any current combatant. It also attacks at its own rate, whether or not the player is attacking. (Horde Minion attacks every 2 seconds or so, about twice as often as a Halberd swing, and turning myself to staff it continued to attack despite my own inability to do so, and the ones that are allowed to go to range will go out and attack likewise)

kevin-10:
The familiar stays with its master, at all times.  It swings (and can miss) only when you melee hit the target.  It will move in range to hit, but only if you are still in its range.  It's kind of weird.

kevin-10:
I attacked a target, and tabbed out, and as i was not hitting it, the familiar never hit it either, so thats proff it only hits when you hit.  Also, they go poof after an hour of you not being near it.

kevin-10:
Also, deletes when you log off (out of world if your in a house or inn).

GriffonSpade:
Dangit, I'm a little confused. Not sure which are things you're making a complaint of how my code works or you're informing me of the OSI behavior? I'll assume they're complaints with my code. Please inform me if I'm incorrect in this assumption.

>Also, deletes when you log off (out of world if your in a house or inn).

That was updated after you tested, I think. They stay in-world now. (I originally planned to check and test that first, but forgot and uploaded first; it's only 2 lines of code that are easy to change either way)

>The familiar stays with its master, at all times. It swings (and can miss) only when you melee hit the target. It will move in range to hit, but only if you are still in its range. It's kind of weird. (OSI behavior or complaint?)

Yes, that's how familiars with !AttacksMastersTarget behave. You need a Death Adder or Vampire Bat to leave master's side and attack things (They are the ones with AttacksMastersTarget). 
Please give feedback on the range they go out to as well, if you can. How far will they go out on OSI? It's currently set to 10 in my code, but 12 or RangePerception(16) are also possibilities for that. (Beyond RangePerception they get teleported back to master). 
Do familiars react differently on OSI? IE, if you flash tab button, (puts you in war mode, then back to peace, nullifying combatant), do they continue attacking? (Note that if you have no combatant and a creature attacks you and is not already aggro'd or been aggro'd by you, you will automatically set it as your combatant)

>I attacked a target, and tabbed out, and as i was not hitting it, the familiar never hit it either, so thats proff it only hits when you hit. Also, they go poof after an hour of you not being near it.

Any time you have a combatant you will attempt to attack them any time they're in range and you're able to attack (ie not frozen or blessed or whatever). When you go into peace mode it actually removes your current combatant, setting it to null. However, the other methods mentioned also work to set your combatant (Double clicking while in war mode, casting a hostile spell on the target, being attacked without any current combatant[ONLY if they have not aggressed you or been aggressed by you already], and attacking with a ranged weapon).

kevin-10:
On OSI, the familiar never leaves your side.  Like I said, the familiar will move to the target as long as it's still within 1 tile of its master.  When the master hits, the familiar will then take a swing at the target.  So, the only time a familiar attacks is if you are attacking the target melee.  Also, I was wrong, as all familiars behave this way.  As for logout, the familiar deletes when you logout from the world, and it after an hour of it not being near you, i.e. When you recall away.  This is the behavior on OSI, not your changes.

So, it looks like we were both wrong.

kevin-10:
So, to match OSI, basically keep the follow functionality the same.  Remove all functionality that makes the familiar attack.  A hook will need to be added at the very end of OnHit method is BaseWeapon.cs, which can point to a static class in BaseFamiliar that handles the familiars action.

GriffonSpade:
>A hook will need to be added at the very end of OnHit method is BaseWeapon.cs, which can point to a static class in BaseFamiliar that handles the familiars action.

No such hook should be necessary. That should be handled by it only attacking master's Combatant. Or do you mean it won't even attack if a player is swinging and missing? Also, does this mean that familiars are never directly targeted ever?

Also, it seems REALLY damned weird for at least the combat familiars to never leave their master's side. But eh, that can be left for custom work.

Do familiars not recall to other maps with their master on OSI? Seems kinda weird if they don't. Do they even teleport to their master at all for any reason?

I can make some changes today, the setup I have is pretty simple, and modifying it this way should just be messing with a couple checks. Then we'll take another look at it.

kevin-10:
OK man, I've written the mechanics on EA 3 times now.  If you refuse to do any testing on EA, and insist on making changes based on memory and opinions, then your pull request won't get approved, its as simple as that.

GriffonSpade:
-Changed Reverted to deleting familiar when master is no longer in the game world (ie internal map after being fully logged out)
-Changed Familiars only attack if both summoner Combatant and familiar are adjacent to summoner

Familiar targeting change should result in them not being targeted, since the master would also be attacking any time the familiar is.

>A hook will need to be added at the very end of OnHit method is BaseWeapon.cs, which can point to a static class in BaseFamiliar that handles the familiars action.

Specifically, this should be covered by how the familiar's combatant is set every AI tick:

@            m.Combatant = GetFamiliarCombatant(master.Combatant, master);
@

@        private IDamageable GetFamiliarCombatant(IDamageable mc, Mobile master)
        {
            if (mc == null)
            {
                return null;
            }

            if (m_Mobile.InRange(master, 1) && master.InRange(mc.Location, 1))
            {
                return mc;
            }

            return null;
        }
@

This results in the familiar's combatant being set to null and simply following the summoner any time:
1) The summoner's combatant is null
2) The summoner's combatant is not within 1 tile of the summoner
3) The summoner's familiar is not within 1 tile of the summoner

PR #2344, opened 2017-07-11
- New Doom Decoration Artifacts (thanks Sareus)

Argalep:


kevin-10:
I would have derived BaseArtifactLight from BaseLight, and simply added an artifact rating property to it.

kevin-10:
See, so much easier!

PR #2439, opened 2017-08-05
Fixs

Argalep:
* BattleWarningGump Exploit fix.
* add HiddenQuestItemHue option
* Virtue Quests – Honor fix.
* GobletOfCelebration fix.
* BagOfJewels fix.
* DeBoorShield fix.


kevin-10:
When is the quest item hue hidden?

Argalep:
I checked osi. Quest item given for quest is in normal color.

kevin-10:
Yea, I see.  Quest item is embedded in Item.cs so there isn't a way around this I suppose.

PR #2449, opened 2017-08-09
* Perfect Timing Quest Fix.

Argalep:
Note: It is necessary to run [GenSutek command after clearing old objects.


kevin-10:
Looks good man!  One question, I see new world generation is handled, but will this auto-generate on existing shards?

Argalep:
It is necessary to run [GenSutek command after clearing old objects.

PR #2458, opened 2017-08-13
Update BladedItemTarget.cs

Hjornskar:
Prevents non-movable / locked down / secured ICarvable items from being carved.


kevin-10:
Thank you!

PR #2502, opened 2017-08-28
Fix about null delegate

Daviex:
If put the CommandRegister into the Initialize, would not compile because of the m_Instance result to be null, so by putting it into the Constructor, it would work anyway because of the constructor be called by Setup.cs with his Initialize, and not give error anymore to compile.

### DIFF
--- a/Scripts/Services/Revamped Dungeons/DespiseRevamped/DespiseController.cs
+++ b/Scripts/Services/Revamped Dungeons/DespiseRevamped/DespiseController.cs
@@ -108,7 +108,6 @@ namespace Server.Engines.Despise
             m_Enabled = true;
             m_Instance = this;          
             
-            //If put this into the Initialize, instance wouldn't exist and give error because m_Instance would be null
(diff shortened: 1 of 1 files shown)


kevin-10:
That won't work, as the command will only be registered when the controller is initially constructed.  It needs to be available for existing shards.  It needs to be pointed to a static method.

Daviex:
You are right, I thought wrongly that this will start the method when the Command into Setup.cs will be Initialized, but it won't until the command is going to be called back. I pushed a new "possible" fix, what you think about this?

kevin-10:
Thanks for the post.  I already fixed it in the Initialize() method.  m_Instnace is set before Initialize() is called, so if m_Instance is not null, the command will be registered.  This command is only for existing shards with the DespiseContoller already existing.  I'm assuming your crash was due to it being a brand new shard, or the Despise Revamped was not setup.

Daviex:
It was not a bug I had, I just readed people complaining and wanted to help find a solution ^^

kevin-10:
Thank you.

PR #2515, opened 2017-09-04
Quest gump

Argalep:
* those whose description is closed are not listed

screenhunter 189


kevin-10:
Is that EA?  If so, our Q&A gumps don't look like that do they?

Argalep:
Yes EA, I changed Q&A gump

kevin-10:
Sweet, thanks.

PR #2559, opened 2017-09-26
NewMobileAnimation Packet fix.

Argalep:
* add AnimationType
* Eat animation fix.
* Gargoyle fly animation fix.
* BaseCreature hurt animation fix.
* salute and bow animation fix.

### DIFF
--- a/Scripts/Items/Consumables/Food.cs
+++ b/Scripts/Items/Consumables/Food.cs
@@ -179,7 +179,7 @@ namespace Server.Items
                 from.PlaySound(Utility.Random(0x3A, 3));
 
                 if (from.Body.IsHuman && !from.Mounted)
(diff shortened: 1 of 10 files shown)


kevin-10:
Very nice work.  I was testing these last night.  Some are for gargs only (you've identified these) and some are for animal/monster types.  The animations such as block, parry, etc, are these used on EA?  Also, we'll need to implement the attack ones.   Each different attack type has its own action number.

Argalep:
yes, block, parry etc. animations are used.

kevin-10:
Ok, will you be working on that? If not, I can do it.

Argalep:
Of course, you can work on it. I will inform you if I have a different information.

PR #2651, opened 2017-10-16
* Termur Region Fix.

Argalep:


kevin-10:
Why is DamageType removed from spellhelper?  That needs to be in there for AOE spells revealing on damage.

kevin-10:
Disregard, I see your fix.

PR #2661, opened 2017-10-19
Fixed imbuing resist weight

kevin-10:


Argalep:
Serialize Problem

Tasanar:
I had to delete all my ancient Smith hammers today after the shard restarted. There is a serious issue with this.

PR #2724, opened 2017-11-11
Update BasePeerless.cs

Hjornskar:
Corrected logic and converted DateTime to TickCount.


kevin-10:
Well done.  Eventually we need to remove that fire ring, as its actually the wild fire spell they would cast.

PR #2782, opened 2017-11-29
PutridHeart Wrong Hue

xG00BERx:
Fixed hue for Putrid Heart.

### DIFF
--- a/Scripts/Items/Quest/PutridHeart.cs
+++ b/Scripts/Items/Quest/PutridHeart.cs
@@ -20,6 +20,7 @@ namespace Server.Items
         {
             Stackable = true;
             Amount = amount;
+            Hue = 2599;
         }
 
         public override void OnDoubleClick(Mobile from)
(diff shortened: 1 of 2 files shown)


xG00BERx:
https://uo.stratics.com/database/view.php?db_content=hunters&id=2490

http://www.uoguide.com/Swoop

Then after seeing that is was different, I logged into my RealUo account and went on test and animal lored it and the one I had lored had roughly 1400 HP and that resists etc were all closer to the Stratics version than they were to the Swoop on ServUO so I figured there was enough reasonable evidence aka 3 sources showing that the data was correct on the website so I changed it and pushed it or whgever.
After that I figured the 3 sources outranked the one on ServUo

Tasanar:
Good research. In the future just post that information with some of these reports.

Tasanar:
Good research. In the future just post that information with some of these reports.

Also those resists are wrong. Even on UOGuide it says for fire 60 - 75, not the 70 you have coded.

Tasanar:
I see stratics also gives slightly different numbers.

Your new hit points are spot on though.

xG00BERx:
So what should I do, I can drop the request and fix the fire, not sure how I missed the extra 5.
The rest of the numbers seem ok with me, what else would you think is needed to be changed?

Tasanar:
Nothing, looks good.

PR #2811, opened 2017-12-07
Update Raptor.cs

Tasanar:
I do not know how to combine all my pull requests into one when editing right from the github website.


Tasanar:
Going to re-open not sure how to simply edit from github website.

PR #2812, opened 2017-12-07
Update HighPlainsBoura.cs

Tasanar:


Scripts/Mobiles/Normal/HighPlainsBoura.cs
```diff
@@ -140,10 +140,14 @@ public override int GetDeathSound()
         public override void OnDeath(Container c)
         {
             base.OnDeath(c);
+            
+            if (Controlled)
```
kevin-10: You already have a check for Controlled that returns, you can take any additional checks out.

Tasanar: https://www.servuo.com/threads/exploit-rare-stuff-drop-and-bonded-pets.8567/

I will update. Which one do you think is preferred?

Tasanar:
Changed

PR #2828, opened 2017-12-13
Implementation of SHA512 for Passwords

Xeroxxx:
SHA512 support

Password-System requires complete rework (I will do this) for proper salting (generating 'random' salt on first run) and better readability. Passwords are automatically converted from SHA1 to SHA512 after logon and next save.

NOTICE: SHA512 is the new default if you replace Config/Accounts.cfg. Change it back to NewCrypt (SHA1) in Accounts.cfg if you're in need of SHA1.

However keep in mind Server <-> Client communication stays unencrypted and is absolutely vulnerable to man in the middle attacks as client encryption is removed. It only protects if data gets stolen.

MD5 is no go and SHA1 is known to be vulnerable to collisions.

Cheers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,7 @@ reportHistory.xml
 *.DotSettings.user
 *.log
 *.bin
+*.lock
+*.ide
+*.ide-shm
+*.ide-wal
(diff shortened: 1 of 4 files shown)


Xeroxxx:

Hey guys, so whats the status here? No wanted? 

Cheers

Tasanar:
I think it is wanted but Kevin (Dexter) also wants some input on it from Vorspire or dmurphy22 (not very active) before the change is added.

PR #2843, opened 2017-12-19
-Converted BaseTool references to ITool for support of sledge hammer …

kevin-10:
-Converted BaseTool references to ITool for support of sledge hammer and smithy hammers, post-AOS.
-Added Smithy Hammer and Sledgh Hammer to loot drops
-Added Void Spawn to Shame Level 3
-Fel/Tram now have their own instance to void pool stats
-Void Pool Gump context menu entry is now accessible anywhere
-Skree AI is now mage
-Fixed kindling weight
-Putrid hearts no longer to go internal map

### DIFF
--- a/Scripts/Items/Addons/BaseAddonContainerDeed.cs
+++ b/Scripts/Items/Addons/BaseAddonContainerDeed.cs
@@ -86,7 +86,7 @@ namespace Server.Items
         }
 
         #region ICraftable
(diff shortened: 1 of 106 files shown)


Tasanar:
Great work! This also fixes the bulk orders? I actually just got 4 bulk orders in a row, as a player. Ill send you the screen shot.

PR #2845, opened 2017-12-19
Update FishingPole.cs

inpoem:
494,495 lines were duplicated 497, 498 lines, now fixed


kevin-10:
I see the issue, its the wrong names. You can't remove them entirely, a bad serialization error will happen at shard start.

PR #2936, opened 2018-01-09
Update SlayerGroup.cs

Argalep:
fix.


Tasanar:
Yeah this commit compiles fine. Sorry about that.

PR #2939, opened 2018-01-10
Update ResourceDrops.cs

Tasanar:
Fixes a nasty exploit where tamed, bonded, creatures can be brought into these regions and exploited non stop flooding the world with these resources.


Tasanar:
https://github.com/ServUO/ServUO/pull/2940

Dexter pushed this one after I pushed me. If his is better you can remove mine, just thought to keep it out of BaseCreature but I am guessing it really does not matter?

PR #2949, opened 2018-01-10
Update BlackthornData.cs

Tasanar:
Fixes an exploit where players can bring a bonded pet into Blackthorns dungeon and kill / rez it over and over, very quickly, to farm blackthorn artifacts,


kevin-10:
Well done.  FYI, you can omit 'true' with victim.Controlled or you can use victim.Controlled == true.

Tasanar:
Oh great I will do that from now on. It looks cleaner. I asked you on ServUO if one was better than the other. So if it is (victim.Controlled) they will just assume that = true? That is how I am understanding it.

kevin-10:
Yes, and to test for a false, you do (!victim.Controlled).

PR #2954, opened 2018-01-11
Bug Fixes

kevin-10:
- Despise Creature AI should now be in line with EA
- VvV steeds can now be rezzed by its master with 0 veterinary skill
- Ronins are now slower
- Frost spiders now spawn the correct 2 hues, per EA
- Fixed retrieve where items where inside that make it non-retrievable
- Arena Non-Participants can no longer teleport into PVP Arena
- Fixed Vela/Cora item Hues
- Davies locker will now give gump for direction placement
- Added Quest Items to Davies Locker entries
- VvV Battle will now attempt to reset after 5 minutes, provided there are enough VvV players logged on
- Fixed static item z axis on shadowguard items (last room)

Dark Guardian Room Fix 
- Added message
- adjusted cooldown timer
- Dark Guardian now spawns proper stuff

### DIFF
--- a/Scripts/Services/Harvest/HarvestSystem.cs
+++ b/Scripts/Services/Harvest/HarvestSystem.cs
@@ -321,8 +321,12 @@ namespace Server.Engines.Harvest
 
             List<Item> atFeet = new List<Item>();
 
+            IPooledEnumerable eable = m.GetItemsInRange(0);
+
             foreach (Item obj in m.GetItemsInRange(0))
                 atFeet.Add(obj);
+
+            eable.Free();
 
             for (int i = 0; i < atFeet.Count; ++i)
             {
@@ -331,6 +335,8 @@ namespace Server.Engines.Harvest
                 if (check.StackWith(m, item, false))
                     return true;
             }
+
+            ColUtility.Free(atFeet);
 
             item.MoveToWorld(m.Location, map);
             return true;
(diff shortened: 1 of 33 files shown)


Scripts/Services/Harvest/HarvestSystem.cs
```diff
@@ -321,9 +321,13 @@ public virtual bool Give(Mobile m, Item item, bool placeAtFeet)
 
             List<Item> atFeet = new List<Item>();
 
+            IPooledEnumerable eable = m.GetItemsInRange(0);
+
             foreach (Item obj in m.GetItemsInRange(0))
```
Argalep: change to 

foreach (Item obj in eable)

Tasanar:
Okay thank you! Goober has been using my OSI account on Atlantic and TC to test so I guess let the farming begin.

kevin-10:
Well, as a shard owner, you can always change it.  I'll definitely be keeping it at how it was on mine.

Tasanar:
Yeah agreed 100% The goal here is just to be OSI accurate! Not make it how I imagine, in my own head, it should be.

kevin-10:
I also have the luxury of using the reasoning of "we don't do everything like EA", as thats been around by staff on the EasyUO shard from day 1 lol.

Well ya, I'm just saying, on your own shard, you don't always have to follow the stupidity of EA.

Tasanar:
Agreed. Should there also be a check to make sure a player GHOST is not triggering numbers?

PR #2958, opened 2018-01-11
Update PetBondingPotion.cs

Tasanar:
Pet Bonding Potion was missing checks to make sure the item was not already used or in a players backpack prior to being used.

This could result in a player selling the potion and then using it right away, if they still have the cursor up.


Tasanar:
Done! All I did here was copy the engraving tool base so maybe that needs it too?

kevin-10:
Yea, basically any type of item that requires a cursor should have it in the OnDoubleClick where the target is initialized, and when the target is actually invoked (OnTarget).

PR #2989, opened 2018-01-14
Bug Fixes

kevin-10:
- Fixed Underworld stairs for new worlds. Existing will need to change it themselves.
- Special Moves that call BaseWeapon OnHit method will only trigger hit effects once, either it be the initial hit, or the subsequent hit if the intial hit fails.
- Arena Stone gumps now close on facet Change
- Paragon chests no longer spawn trammel t-maps on siege rulesets
- Crystal portals no longer send folks to trammel
- Trash barrel and chests now have a 1% chance to drop items to cavern of discarded
- Fixed bug where chicken coops weren't setting creatures IsStabled to true upon server Load
- Tyballs Shadow now spawns wearing SHroud of the Condemned
- Vice Vs Virtue has some combat changes, per EA.  If a non-VvV player does a beneficial or aggressive action on a VvV player (in fel), they become a VvV combatant (orange) for 30 minutes. Also, if they don't leave the VvV battle zone after a minute, or close the gump without teleporting to a moongate, they become flagged for 30 minutes.
- VvV battle changed to 30 mintes
- Fixed an issue where VvV battles didn't auto Reset
- Fixed an issue where Dard Guardian Room wan't spawning the proper a
- Dark guardians now drop the proper amount of daemon bones
- Fixed Bonus Resources leaking to internal map
- Magincia Housing Lotto now checks (server start) for offending plots not converted to public house placement
- Spells/Targets are canceled prior to Arena Duels beginning
- Added [ResetArenaStats command. Target a arena stone to reset the arena stats
- Fixed issue where golden compase wouldn't decay after server is restarted
- Players are now paralyzed for 2 seconds and given a strict warning before entering the hallway of death
- Fixed issue where players were getting spirituality when not healing (ie hits were already at max)
- Field spells and wall of stone cast range increased to 15 (TOL Check)
- Mana shield no longer reveals every tick
- Arcane circle strength now depends on Spellweaving mastery level, per EA
- Summon Reaper now per EA

### DIFF
--- a/Data/Decoration/Stygian Abyss/Ter Mur/Underworld.cfg
+++ b/Data/Decoration/Stygian Abyss/Ter Mur/Underworld.cfg
@@ -61,12 +61,14 @@ SecretShadowWallNS 0x363A
 1136 1166 -12
 
 # Secret Passage: Stone stairs
-Static 0x0752 (Hue=2000)
+Static 0x3F0 (Hue=1945)
 1139 1165 -27
 1138 1166 -22
+1138 1165 -22
 1139 1166 -27
 1140 1165 -32
 1137 1166 -17
+1137 1165 -17
 
 # Secret Dungeon Wall
 SecretDungeonWallNS 0x0242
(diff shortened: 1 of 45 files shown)


kevin-10:
Good catch.

Tasanar:
Other than that looks great! Thanks for all the hard work.

PR #3005, opened 2018-01-16
Bug Fixes

kevin-10:
- Fixed issue with VvV notoriety - report back!
- Guardzone/Guardwhacking should now be per EA - report back!
- Added much need IsMonster propery to BaseCreature.cs
- Added pack checks for gorgon lense
- remove durability from Blackthorn Epaulettes and Hooded robes. Do we need to do this to the dresses too?
- Changed GetMobilesInRange() that appear as teh loop iterator to a variable so Free() is called. Still need to implement for GetItemsInRange() which shouldn't be as tedious.

### DIFF
--- a/Scripts/Spells/Spellweaving/ArcaneCircle.cs
+++ b/Scripts/Spells/Spellweaving/ArcaneCircle.cs
@@ -147,7 +147,7 @@ namespace Server.Spells.Spellweaving
 
             //OSI Verified: Even enemies/combatants count
             IPooledEnumerable eable = Caster.GetMobilesInRange(1);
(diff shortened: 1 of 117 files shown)


Scripts/Spells/Spellweaving/ArcaneCircle.cs
```diff
@@ -146,6 +146,7 @@ private List<Mobile> GetArcanists()
             weavers.Add(Caster);
 
             //OSI Verified: Even enemies/combatants count
+            IPooledEnumerable eable = Caster.GetMobilesInRange(1);
             foreach (Mobile m in Caster.GetMobilesInRange(1))	//Range verified as 1
```
Argalep: foreach (Mobile m in eable)

Issue #3040, opened 2018-01-21
Skull Rug, Dolphin Rug, Rose Rug


BugBot:
These all spawn loot type regular on ServUO but are Blessed on real UO (confirmed)
[ATTACH=full]10748[/ATTACH]

So I see there are two styles but I do wonder HOW they are actually set up.

Also according to UO Wiki there should be two sizes, not sure how to confirm.
[QUOTE]  Oval Rug
[LIST]
[*]Choice of 3 styles [URL='http://uo.com/wp-content/uploads/2015/10/rugs_south.jpg']pictured here[/URL]
[LIST]
[*]The skull and cross bones will give 1 random Treasure chest map each week.
[*]The Rose Rug will give one random seed each week.
[*]The Dolphin will give one random MIB each week.
[/LIST]
[*]Choice of 2 sizes, 7 x 7 or 3 x 5
[*]East or South facing
[/LIST]
[/QUOTE]
[URL]https://uo.com/wiki/ultima-online-wiki/items/veteran-rewards/[/URL]

Reported By: Tasanar
Report Thread: http://servuo.com/threads/skull-rug-dolphin-rug-rose-rug.8848/

PR #3049, opened 2018-01-23
Bug Fixs.

Argalep:
* Vet Reward rugs (dolphin, skull, rose) fix. #3040
* Undertaker's Staff crash fix. #3050

### DIFF
--- a/Scripts/Gumps/RugGump.cs
+++ b/Scripts/Gumps/RugGump.cs
@@ -1,0 +1,76 @@
+using System;
+using Server.Items;
+using Server.Network;
+
+namespace Server.Gumps
+{
+    public class RugGump : Gump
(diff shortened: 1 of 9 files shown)


kevin-10:
-Regarding the rugs, where were you when we discussed on the forums!  Is this gump EA accurate?  
-There has to be a better way than changing the access level of BaseAddonDeed Target
-You have the south/east facing backwards. It was backwards before, and I fixed it
-All existing deeds will delete now because they derive from the BaseRugAddonDeed

Tasanar:
I like the way you did it Dexter. That could have been pre publish.

kevin-10:
Well, its a vet reward, so why wouldn't we use the reward option gump?

PR #3051, opened 2018-01-23
fix.

Argalep:


Tasanar:
https://uo.com/wiki/ultima-online-wiki/items/veteran-rewards/

I would 1. Add a check for IsRewardItem and also veteran rewards, including these, can be used by anyone.

Tasanar:
"All rewards can only be claimed by accounts of the required age, but most can be traded to, and used by younger accounts. This does not apply to ‘account bound’ rewards such as the red soulstone and shard shields."

Taken from official UO wiki.

Argalep:
http://www.xn--28j2a0gvbiy.com/entry/2012/11/26/121008

image

The picture also says in the message of use:

"Your account is not old enough to use this item. Months until you can use this item : 8"

Tasanar:
That is older. They removed all age restrictions except for soul stones and shard shields.

Tasanar:
http://www.uoguide.com/Publish_83
https://www.servuo.com/threads/veteran-rewards-no-longer-require-a-set-age-to-use-just-to-claim.4060/

"*Removed account age limit to use vet reward items. This change does not remove the account bound limitation set on some vet rewards."

Tasanar:
Those pictures are from when the rugs first came out. They have completely changed since then.

kevin-10:
Who should be able to take the item then?  I would say anyone on the owner account and/or anyone on the account that placed the addon.

Tasanar:
I would say house owner. There is a chance these have an access level any anyone with that access level can take.

Argalep:
When you convert the addon to deed, will not write veteran reward on deed

Issue #2916, opened 2018-01-06
Makefile fuzzy


Xeroxxx:
https://github.com/ServUO/ServUO/blob/53fb7a4253a373284a6d484d1b2fcbfb082e0da8/Makefile#L11

In which Linux distribution is this supposed to work?

I'm always getting mono64 missing and have to edit this manually...as Debian, CentOS, Ubuntu, Arch do not work like that.

Xeroxxx:
Fixed

PR #3063, opened 2018-01-25
Update Makefile | Monofix

Xeroxxx:
Updating Makefile so mono64 is used on Mac OS X and not ALL Unix/Linux based systems. It stays 'mono' on Linux Distributions even 64 Bit.


Xeroxxx:
#2916

PR #3064, opened 2018-01-25
Improved Startup Output | Added help argument

Xeroxxx:
Added -h and -help argument to ServUO

Improved Startup coloring

Added Text to differ between Debug and Release Builds

Added Output with which .NET TargetVersion the Executable is compiled.


Xeroxxx:
servuo com-version

PR #3066, opened 2018-01-25
Synchronizing BuildOptions

Xeroxxx:
Synchronizing BuildOptions so Compile.WIN.bat does the same as compiling with Visual Studio and viceversa


Xeroxxx:
Basiclly the decision between optimize builds with debug or not. So this must be changed in bat and vs solution.

Issue #3055, opened 2018-01-24
Staff Member Online or not - dead giveaway


BugBot:
So this may not be a bug and there really is no way to test, except I did go on TC and it simply told me a Game Master would respond as soon as possible and said nothing about no one being online.

[ATTACH=full]10778[/ATTACH]

Players will send a message to SEE if a staff is on. When they get that they will then attempt to afk farm, macro, ect. It is a dead give away.

[code]if (!isStaffOnline)
            {
                entry.Sender.SendMessage(
                    "We are sorry, but no staff members are currently available to assist you.  Your page will remain in the queue until one becomes available, or until you cancel it manually.");
            }[/code]

I know I can just delete that, just wanted to point it out as it seems custom and can be used against a shard more than anything.

Reported By: Tasanar
Report Thread: http://servuo.com/threads/staff-member-online-or-not-dead-giveaway.8865/

PR #3067, opened 2018-01-25
Update Page Help System | Added possibility to configure

Xeroxxx:
Added option to adjust website and support email address

Added support to hide the message that no staff member is online


Xeroxxx:
Issue #3055

Xeroxxx:
Changed behavior in HelpGump.cs, making use of new SupportWebsite Feature in Help System.

Last uo.com reference is in HelpGump.cs Line 106 making use of a short name. I will think about this in future maybe just shorten the SupportWebsite string for this text.

However this is now complete from my side an can be merged.

PR #3072, opened 2018-01-26
Bug Fixes

kevin-10:
Runic Reforging
- fixed but where resists weren't being applied to Armor
- name#1 and name#2 now appear randomly as the prefix/suffix
- updated prefix/suffix specific property count
- Random properties (not from prefix/suffix) are now randomly chosen from pool of unused prefix/suffix lists. It used to use standard RunicProperty. We need to check how standard loot is handled
- Fundamental and/or Structural options now hue the item
- Removed BlockRepair from BaseArmor/BaseWeapon/BaseJewel/BaseClothing as negative attribute NoBlock made this redundant

Bug Fixes
- All t-maps now reset, regardless of Facet
- Added lockdown owner to access of rug resources
- Corrosive slimes should now drop the prop peerless keys
- Change City Loyalty Electoin DateTime.UtcNow to DateTime.Now so election times conincside with server time
- Added Savage Mask and Orcish Kin Mask to imbue list
- Reduced VvV Statloss to 5 minutes

### DIFF
--- a/Scripts/Items/Addons/DolphinRug.cs
+++ b/Scripts/Items/Addons/DolphinRug.cs
@@ -122,7 +122,7 @@ namespace Server.Items
         {
             BaseHouse house = BaseHouse.FindHouseAt(from);
 
-            if (house != null && house.IsOwner(from))
+            if (house != null && (house.IsOwner(from) || (house.LockDowns.ContainsKey(this) && house.LockDowns[this] == from)))
             {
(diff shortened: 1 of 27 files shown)


Tasanar:
Very curious to test this and see the different loot combinations it comes up with. Hope nothing too OP! 

Great publish.

kevin-10:
Its not loot (except the bug where resists weren't appearing on armor), the fixes should only show on reforging.

PR #3084, opened 2018-01-27
Adjustment for proper output of Environment.OSVersion

Xeroxxx:
Microsoft keeps output of 6.2.x for compatibility reasons starting Windows Server 2012 and 8.

Adding manifest file solves this issue.

version

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,4 +28,3 @@ reportHistory.xml
 *.log
 *.bin
 .vs/
-/ServUO.exe.manifest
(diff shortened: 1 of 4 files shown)


Xeroxxx:
Manifesting the .exe for Windows 8.1 or Windows 10 will not have any impact when run on previous operating systems

https://msdn.microsoft.com/library/windows/desktop/dn481241(v=vs.85).aspx?f=255&MSPPError=-2147217396

PR #3099, opened 2018-01-30
Update BankerGump.cs

timginter:
Check if depositing gold/checks was successful when withdrawing from secure account.
Bug when not using account gold system - gold from secured account was removed even if depositing gold/checks failed (e.g. full bank).

### DIFF
--- a/Scripts/Gumps/BankerGump.cs
+++ b/Scripts/Gumps/BankerGump.cs
@@ -149,10 +149,15 @@ namespace Server.Gumps
                                         from.SendLocalizedMessage(1155867); // The amount entered is invalid. Verify that there are sufficient funds to complete this transaction.
                                     else
                                     {
-                                        Banker.Deposit(from, v, true);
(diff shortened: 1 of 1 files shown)


kevin-10:
If new gold system is disabled, shouldn't secure account be disabled too?

timginter:
Not sure if it has to. If account currency disabled would mean a pre TOL serwer then probably both secure account and Wall Safe.

BankerGump doesn't have a check if account currency is enabled, though, and Wall Safe still works with account currency disabled. You can still deposit into Secure Account from bank boxes with account currency disabled and use that gold in Wall Safes

kevin-10:
Right, but the a bank gump should have a TOL check. That would solve the problem. Also, your change would be ok in case some one wants custom features.

timginter:
Pardon, my bad - didn't notice BankerGump entry has a TOL check separate from Account Currency enabled/disabled...

Closing

PR #3106, opened 2018-02-01
Bug Fixes

kevin-10:
- Removed Underworld cannons from decoration, they will now be handled in the Underworld Generate so we can add the cannoneers.
- Added cannoneers to the underworld cannons
- Fixed equip last weapon, as well as crash
- Added FindItem<T> and FindMobile<T> method as a Map extension some may find very useful
- Base Hires fixed. They now count as 2 stable slots, and show the proper property when hired.
- Hires now show the proper commands
- Fixed the itemID and hue of Gray Goblins (Thank you Tauron)
- Updated Solen Hive Range (fresh shards only)
- Added version flag for future use to SpawnerPersistence, so we don't skip over versions for those who don't update regularly.


Xeroxxx:
Finally :) and fresh objects.xml 

- Updated Generation Categorization
- Fixed GenCategorization crash Fix

Thx

PR #3126, opened 2018-02-06
Update Honor.cs

Tasanar:
Missing various system messages for the Honor virtue.
https://www.servuo.com/threads/honor-virtue-missing-a-few-messages.8962/

### DIFF
--- a/Scripts/Services/Virtues/Honor.cs
+++ b/Scripts/Services/Virtues/Honor.cs
@@ -111,8 +111,10 @@ namespace Server
             if (honorTarget.ReceivedHonorContext != null)
             {
                 if (honorTarget.ReceivedHonorContext.Source == source)
+                {
+                    source.SendLocalizedMessage(1115882); // You don't need to declare again. You are already under Honorable Combat with this target.
                     return;
-
+                }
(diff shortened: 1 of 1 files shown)


kevin-10:
Whats up withthe Travis CI errors?

Tasanar:
I have NO idea what any of that means.

All I did was add in a few lines, that I already applied to Heritage, for the missing system messages.

Tasanar:
When you get home take a look from your computer. I already applied it to Heritage and it works great. When I looked at the errors it also had nothing to do with Honor, the only script I edited. Maybe because I make my commits by editing the script through GitHub? Not actually uploading anything? No idea what TravisCI is or what it does.

Xeroxxx:
The travisci errors happening for month. The are showing up because I submitted the fix and its working again. So it switches from permanent error to failed. No worries it still compiles, just a case sensitive issue.

Xeroxxx:
Sp it gets deactivated if there are too many fails, no it works basiclly again and starts being displayed.

Xeroxxx:
@kevin-10 I will message you on discord tomorrow.

PR #3161, opened 2018-02-17
Honesty Generation

Vorspire:
Improvements and features for large area spawn generation, first implemented by Honesty Virtue.


Xeroxxx:
I'm not the one asked to review. Just wanna add that the build checks fail for the "pushes" individually as the branch "Honesty" is with old project files that fail on travis ci. As you can see the PR succeeds.

Well done

kevin-10:
Is it necessary to run SpawnArea.Instantiate every server start?

Vorspire:
Regeneration is necessary if things like regions change, or the map files update.
If we know all of the factors taken into account when generating certain spawn areas, we could implement a checksum of sorts to skip regeneration and load a set of saved points. Or is that overkill? :D

PR #3180, opened 2018-02-24
Update BaseMount.cs

Argalep:
Console.Write clear

### DIFF
--- a/Scripts/Mobiles/Normal/BaseMount.cs
+++ b/Scripts/Mobiles/Normal/BaseMount.cs
@@ -529,7 +529,6 @@ namespace Server.Mobiles
             {
                 if (m_Type == BlockMountType.RidingSwipe)
                 {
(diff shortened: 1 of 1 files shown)


Xeroxxx:
Looks good. But I've question thats unrelated to removing debug output.

Why can't the lines starting 532 be simplified to

`           public bool IsExpired(BaseMount mount)
            {
                if (m_Type == BlockMountType.RidingSwipe)
                {
                    if (DateTime.UtcNow < m_Expiration)
                    {
                        return false;
                    }
                    else if (m_Mount != null && mount != null && m_Mount == mount && mount.Hits >= mount.HitsMax)
                    {
                            BaseMount.ExpireMountPrevention(m_Mobile);
                            return true;
                    }
                    else if (m_Mobile != null && m_Mobile.Hits >= m_Mobile.HitsMax)
                    {
                            BaseMount.ExpireMountPrevention(m_Mobile);
                            return true;
                    }
                    return false;
                }`

PR #3184, opened 2018-02-26
Add item decay config

dariys:
Add item decay tooltip
Add config for item decay timer
Add virtual property to allow derived items to specify a multiple of the default decay time

### DIFF
--- a/Config/General.cfg
+++ b/Config/General.cfg
@@ -12,3 +12,6 @@ ShowStaffOffline=true
 SupportEmail=
 #String Syntax: https://support.uo.com
 SupportWebsite=
+
+# Time in minutes that generic items take to decay when on the ground
+DefaultItemDecayTime=60
--- a/Server/Item.cs
+++ b/Server/Item.cs
@@ -2012,20 +2012,34 @@ namespace Server
             get { return false; }
         }
 
-        private static TimeSpan m_DDT = TimeSpan.FromHours(1.0);
+        private static TimeSpan m_DDT = TimeSpan.FromMinutes(Config.Get("General.DefaultItemDecayTime", 60));
 
+        public virtual int DecayMultiplier { get { return 1; } }
+
+        [CommandProperty(AccessLevel.GameMaster)]
         public static TimeSpan DefaultDecayTime { get { return m_DDT; } set { m_DDT = value; } }
 
-        [CommandProperty(AccessLevel.Decorator)]
-        public virtual TimeSpan DecayTime { get { return m_DDT; } }
+        [CommandProperty(AccessLevel.GameMaster)]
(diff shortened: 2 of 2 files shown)


Xeroxxx:
I would like to see the options in General.cfg, not worth it adding separate configfile for two options. We already have a shit ton of configs :D

kevin-10:
Are there any items that show the decay time for actual decay?  If not, I don't see a need to make core edits for custom code.

dariys:
@Xeroxxx - I'll move the config options into General.cfg

@kevin-10 - I'm adding an option here to display decay via tooltip to users.  If this is enabled (off by default, to match standard behavior) - when hovering over items, it shows in the tooltip for items (i.e. stuff on the ground)

dariys:
@Xeroxxx - Looking at the options in General.cfg though - these really don't seem like they fit there...

kevin-10:
I get that, but like I said, I don't see a need to make core edits for custom code. Look at BaseDecayingItem.cs if you want to use this functionality.

dariys:
@kevin-10 Just to be clear - this is for all items that decay from the world - i.e. dropped on the ground - piles of ore, logs, whatever.  This change makes the decay time configurable (from the default of one hour), and adds the configuration to display that time on the tooltip of the item (again - only when the item is on the ground.  When it is in the bag, there is no decay display on the tooltip).  The only piece of code that supports custom edits is the decay multiplier - which for example - I have ore set to a higher multiplier so it lasts longer.  If that is the piece you object to - I can remove that.

Xeroxxx:
Well, I mainly follow @kevin-10 opinion. IMHO making DecayTimer configurable is useful, without adding the tooltip. However everything fits in General.cfg, doesn't matter whats in there, because its "General".

kevin-10:
@dariys I'm ok with the decay modifier, just don't think we need the addition to the object property list.

dariys:
@kevin-10 Even as a configurable option?  It's defaulting to false (not shown) with what I've submitted here.

Xeroxxx:
@Tasanar 
Configurable DecayTimer and Multiplier is still a core modification with or w/o tooltip.

I agree with @kevin-10 configurable DecayTimer and Multiplier is ok. I pledge for moving it in general.cfg no need for new configs. Tooltip is not necessary.

Tasanar:
I know, I just meant the tooltip edits.

kevin-10:
Agreed. @dariys, just remove the tooltip (object property stuff) but the config stuff is ok for decay boolean and decay multiplier.

dariys:
Updated

- config is now in general.cfg
- Only decay time & decay override mechanics included - tooltip addition is removed

PR #3186, opened 2018-02-28
Salvage bag fix

dariys:
Since the salvage method itself just looks for items that are IScissorable, the additonal checks to determine if there are anything to salvage just prevent starting salvaging when there are only items that don't meet them - specifically things like Oil Cloth.

This change allows the salvage cloth option to be present when there are any IScissorable items in the salvage bag.

```
        private void SalvageCloth(Mobile from)
        {
               .....

                Item item = Scissorables[i];
                if (item is IScissorable)
                {
                    if (((IScissorable)item).Scissor(from, scissors))
                    {
                        salvaged++;
                    }
```


kevin-10:
Thank you for the contribution.

PR #3191, opened 2018-03-02
Starting to fix Oil Flasks

Tasanar:
Please review. 

https://www.servuo.com/threads/mechanical-pet-leather-wolves.9066/

FlaskOfOil was not used. I deleted it since the item is called an OilFlask

I also believe it makes more sense to be in Resources.


kevin-10:
Was it not used at all?  If it was, you can use TypeAlias in the new oil flask so any existing won't be deleted:

[TypeAlias("Server.Items.FlaskOfOil ")]
public class OilFlask: Item
{

Tasanar:
No it was not. If you look it was only referenced once in ModifiedClockworkAssembly but was commented out. It never naturally spawned anywhere or did anything.

kevin-10:
Yea but other shards may have spawned them.

Tasanar:
Done sir.

PR #3216, opened 2018-03-14
Update for pets skill decrease on res

timginter:
Pets lose 0.1% of each skill when resurrected by any player, not only ControlMaster. 0.2% when resurrected by NPC


timginter:
Pardon, NPC resurrect is handled somewhere else. Have to test if other player resurrecting someone else's pet ends in a 0.2% skill penalty

Issue #3204, opened 2018-03-06
Vendor Search Bug


BugBot:
Go into a town, click on Vendor Search, and leave town with the menu up.

You can go into other facets, like Ilshenar, and still search vendors. You can also create maps and teleport out of these areas.

On real UO the second you recall, use a teleporter, or a moongate, all vendor search menus go away.

IF when you try to do anything within the menu, search, ect, it will give you this message if not in a guarded town or house.
[ATTACH=full]11051[/ATTACH]

Also it "acts" like a recall spell but should not actually check for LRC or reagents.

Reported By: Tasanar
Report Thread: http://servuo.com/threads/vendor-search-bug.9088/

PR #3228, opened 2018-03-20
Validate vendor search use upon gump response

DarkLotus:
Add a new validation check in the gump response handler for any button but exit. not OSI perfect as gump stays open when you leave a safe area, but solves the exploit potential re #3204


Tasanar:
The problem is here:

https://www.servuo.com/threads/vendor-search-bug.9088/

Tested on EA vs ServUO. Much more open to exploit on ServUO as EA will prevent you from using the menu, outside of town.

PR #3235, opened 2018-03-22
fixs.

Argalep:
* VvV Chat notification fix. (add MessageLocalizedAffix EC Support)
* the position of the items in the bag of the deceased character is deleted.
* EC Spell Cast Crash Fix.

### DIFF
--- a/Scripts/Items/Consumables/BankCheck.cs
+++ b/Scripts/Items/Consumables/BankCheck.cs
@@ -213,7 +213,7 @@ namespace Server.Items
 					"",
 					AffixType.Append,
 					String.Concat(" ", m_Worth.ToString()),
-					"")); // A bank check:
+					"", false)); // A bank check:
 		}
 
 		public override void OnDoubleClick(Mobile from)
--- a/Scripts/Items/Consumables/TreasureMap.cs
+++ b/Scripts/Items/Consumables/TreasureMap.cs
@@ -897,7 +897,7 @@ namespace Server.Items
                         "",
                         AffixType.Append,
                         String.Format(" completed by {0}", m_CompletedBy == null ? "someone" : m_CompletedBy.Name),
-                        ""));
+                        "", false));
             }
             else if (m_Decoder != null)
             {
--- a/Scripts/Items/Tools/Spyglass.cs
+++ b/Scripts/Items/Tools/Spyglass.cs
@@ -41,8 +41,8 @@ namespace Server.Items
         {
             from.LocalOverheadMessage(MessageType.Regular, 0x3B2, 1008155); // You peer into the heavens, seeking the moons...
 
-            from.Send(new MessageLocalizedAffix(from.Serial, from.Body, MessageType.Regular, 0x3B2, 3, 1008146 + (int)Clock.GetMoonPhase(Map.Trammel, from.X, from.Y), "", AffixType.Prepend, "Trammel : ", ""));
-            from.Send(new MessageLocalizedAffix(from.Serial, from.Body, MessageType.Regular, 0x3B2, 3, 1008146 + (int)Clock.GetMoonPhase(Map.Felucca, from.X, from.Y), "", AffixType.Prepend, "Felucca : ", ""));
+            from.Send(new MessageLocalizedAffix(from.Serial, from.Body, MessageType.Regular, 0x3B2, 3, 1008146 + (int)Clock.GetMoonPhase(Map.Trammel, from.X, from.Y), "", AffixType.Prepend, "Trammel : ", "", false));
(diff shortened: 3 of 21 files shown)


Argalep:
I wanted to disrupt the commonly used functions. When I check, they are not a problem for the EC client. I thought we could use SendLocalizedMessageECSupport in messages to be affixed.

Vorspire:
More named methods just adds bloat, why can't the normal methods just be changed to detect EC capability on the netstate the packet is being sent to, like everything else?

kevin-10:
This packet is called from Mobile.cs so passing the netstate variable in would be easy.

kevin-10:
@Vorspire , the problem is for the way this packet is sent some of the times, is when the packet is sent its set static and sent to all the looped net states in range (IE PublicOverheadMessage), therefore passing in a net state argument wouldn't be possible at that point.  Maybe set a  separate packet for the EC clients (would be null until a EC client is found in the loop):

`public void PublicOverheadMessage(
			MessageType type, int hue, int number, AffixType affixType, string affix, string args, bool noLineOfSight)
		{
			if (m_Map != null)
			{
				Packet ccp = Packet.Acquire(new MessageLocalizedAffix(m_Serial, Body, type, hue, 3, number, Name, affixType, affix, args));
				Packet ecp = null;

				var eable = m_Map.GetClientsInRange(m_Location);

				foreach (NetState state in eable)
				{
					if (state.Mobile.CanSee(this) && (noLineOfSight || state.Mobile.InLOS(this)))
					{
						if(state.IsEnhancedClient)
						{
							if(ecp == null)
							{
								ecp = Packet.Acquire(new MessageLocalizedAffix(state, m_Serial, Body, type, hue, 3, number, Name, affixType, affix, args));
							}
						}
						else
						{
							state.Send(ccp);
						}
					}
				}

				Packet.Release(p);

				eable.Free();
			}
		}`

Also, create an overload MessageLocalizedAffix constructor to avoid editing tons of calls, and pass in the netstate arg where issues arrise in the future. EC functionality shouldn't be forced unless its for an EC client:

`public MessageLocalizedAffix(
				Serial serial,
-            	int graphic,
-            	MessageType messageType,
-            	int hue,
-            	int font,
-            	int number,
-            	string name,
-            	AffixType affixType,
-            	string affix,
-            	string args)
-            	: this(null, 
					serial, 
					graphic, 
					messageType, 
					hue, 
					font, 
					number, 
					name, 
					affixType, 
					affix, a
					rgs)
-        	{
			}
			
			public MessageLocalizedAffix(
				NetState state,
				Serial serial,
-            	int graphic,
-            	MessageType messageType,
-            	int hue,
-            	int font,
-            	int number,
-            	string name,
-            	AffixType affixType,
-            	string affix,
-            	string args)
-            	: base(0xCC)
-        	{
				if (name == null)
-            	{
-                	name = "";
-            	}
	-           if (affix == null)
	-           {
	-           	affix = "";
	-           }
	-           if (args == null)
	-           {
	-               args = "";
	-           }
	-
	-           if (hue == 0)
	-           {
	-           	hue = 0x3B2;
	-           }
-
-            	EnsureCapacity(52 + affix.Length + (args.Length * 2));
-
-            	m_Stream.Write(serial);
-            	m_Stream.Write((short)graphic);
-            	m_Stream.Write((byte)messageType);
-            	m_Stream.Write((short)hue);
-            	m_Stream.Write((short)font);
-            	m_Stream.Write(number);
-            	m_Stream.Write((byte)affixType);
-            	m_Stream.WriteAsciiFixed(name, 30);
-            	m_Stream.WriteAsciiNull(affix);
 
-            	if (state != null && state.IsEnhancedClient)
				{
					m_Stream.WriteLittleUniNull(args);
				}
				else
				{
					m_Stream.WriteBigUniNull(args);					
				}
			}`

Vorspire:
I would just implement the two different packets, as you say, instantiate them as they are needed. This is how caching is done for features like delta queue processing where you're sending a lot of information to multiple different clients. You don't have to pass a NetState argument because Mobile has a NetState property reference anyway.

Define Packets; p, p_ec;
foreach client, check EC support, instantiate appropriate packet, send it
end loop, release both p and p_ec.

Vorspire:
It's also worth noting that things like World.Broadcast do not call SendMessage (for example's sake), they re-implement SendMessage in a way that doesn't result in one packet being sent to everyone online. Anything that doesn't do this and calls MessageLocalized* or SendMessage, in a loop, is going to have more overhead.

Tasanar:
While the EC spell crash fix looks good I am not sure about the other stuff. Above my level.

kevin-10:
@Argalep can you close this, and just push through the EC spell crash fix. I'll have a fix for the localized message so we're not having to edit other files to accommodate it.  Thanks!

PR #3251, opened 2018-03-26
Update for Mono

Xeroxxx:
Removing "Compile.MONO.bat" there is no need to maintain this or use Mono on Windows.

Updated the Makefile to be inline with Windows "bat" files and Visual Studio.

There is no need to build the core with -debug for production use.

-debug is still an option for ServUO.MONO.exe -debug to enable script debugging


Xeroxxx:
https://www.servuo.com/threads/i-been-unable-to-install-server.7924/

Xeroxxx:
Will be merged/implemented with binary compatibility patch

PR #3253, opened 2018-03-27
Fixs.

Argalep:


kevin-10:
I am reverting this fix. I didn't notice the core edit to Mobile.cs. You are essentially moving all resists down one, which will cause major problems. Making Core edits for standalone systems is a bad idea. Please implement a way for the sphinix buffs without making edits to any existing code such as Mobile.cs and the AOSAttribute.

PR #3261, opened 2018-03-27
fix.

Argalep:
* ResistanceType order fix.
* SphynxFortune duplicate message fix.


kevin-10:
There needs to be a better way of implementing the sphynx buffs without editing the enums.

Argalep:
I thought it was a good idea. I'll be glad, if you help me.

Tasanar:
Is the client hard coded against having so many resists?

At this stage I don't know if UO would add more resists. Though we do have Chaos damage, ect.

If None is at the end it just allows us to reference it without doing anything?

PR #3267, opened 2018-03-28
* add Sphynx Fortune

Argalep:


Scripts/Services/Malas/Sphynx/SphynxFortune.cs
```diff
@@ -218,37 +213,41 @@ public static void StartTimer()
...
+                reader =>
+                {
+                    int version = reader.ReadInt();
 
-            for (int i = reader.ReadInt(); i > 0; i--)
-            {
-                Fountains.Add(new SphynxFortuneArray { Mobile = reader.ReadMobile(), Date = reader.ReadDateTime(), Type = (EnumType)Enum.Parse(typeof(EnumType), reader.ReadString()), TypeValue = (EnumTypeValue)Enum.Parse(typeof(EnumTypeValue), reader.ReadString()), Value = reader.ReadInt() });
-            }
+                    for (int i = reader.ReadInt(); i > 0; i--)
+                    {
```
kevin-10: Mobile = reader.ReadMobile()

If this mobile is null, will this cause any adverse effects?  players that are deleted prior to server restarted will be null upon server start.

PR #3284, opened 2018-04-02
Update Brightblade.cs

Tasanar:
Confirmed on real UO - This artifact should spawn with 255 durability.


Tasanar:
All the mining elementals, except the Dull Copper Elemental, are meant to drop 25 of their respected resources. Confirmed on real UO.

Also dull copper elementals do not drop the crystalline black rock. They are not considered "miners" elementals since they always spawned in the world. (I know shadow and golden do too but they still drop 25 each and the rare resources. Confirmed on real UO) They are still the only elemental that only drops 2 of their respected ore. All others drop 25.

PR #3292, opened 2018-04-03
Bug fixes

kevin-10:
- Fixed Tinker bods
- Added Energy pick for crimson/plat drakes
- Removed conflicting layer from Merchant Crew
- Fixed spells that cannot stack damage
- Fixed Stack Overflow
- Failed Pet Special Abilities and Area Effects no longer use mana
- Void Pool Context Menu selection is only available if you've visited teh void pool

### DIFF
--- a/Scripts/Services/BulkOrders/SmallBODs/SmallTinkerBOD.cs
+++ b/Scripts/Services/BulkOrders/SmallBODs/SmallTinkerBOD.cs
@@ -195,7 +195,6 @@ namespace Server.Engines.BulkOrders
             new Type[] { typeof(Clock), typeof(ClockRight), typeof(ClockLeft) },
             new Type[] { typeof(GoldRing), typeof(SilverRing) },
             new Type[] { typeof(GoldBracelet), typeof(SilverBracelet) },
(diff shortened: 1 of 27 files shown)


Scripts/Services/BulkOrders/SmallBODs/SmallTinkerBOD.cs
```diff
@@ -191,6 +195,8 @@ public override bool CheckType(Type itemType)
             new Type[] { typeof(Clock), typeof(ClockRight), typeof(ClockLeft) },
             new Type[] { typeof(GoldRing), typeof(SilverRing) },
             new Type[] { typeof(GoldBracelet), typeof(SilverBracelet) },
+            new Type[] { typeof(GoldBracelet), typeof(SilverBracelet) },
```
Argalep: duplicate ?

kevin-10: Good catch.

kevin-10:
No, a slight change in teh method where the creature chooses a dispel spell.

Tasanar:
So it could potentially still crash

PR #3338, opened 2018-04-16
+ Region entity enumeration optimizations;

Vorspire:
* Removed the need to instantiate a new list for iteration.
* Checking entity location is less expensive than finding their region.
* Removed Parallel Task implementation and thread safety locks.
* Extra overloads for providing enumeration predicates.

### DIFF
--- a/Server/Region.cs
+++ b/Server/Region.cs
@@ -468,241 +468,192 @@ namespace Server
 			return false;
 		}
 
+		#region Entity Enumeration
 		public List<Mobile> GetPlayers()
 		{
-			var list = new List<Mobile>();
+			return GetPlayers(null);
+		}
 
-			if (m_Sectors != null)
+		public List<Mobile> GetPlayers(Func<Mobile, bool> predicate)
+		{
+			return GetEnumeratedPlayers(predicate).ToList();
+		}
+
+		public IEnumerable<Mobile> GetEnumeratedPlayers()
+		{
+			return GetEnumeratedPlayers(null);
+		}
+
+		public IEnumerable<Mobile> GetEnumeratedPlayers(Func<Mobile, bool> predicate)
+		{
+			if (Sectors != null)
 			{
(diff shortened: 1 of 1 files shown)


Tasanar:
So if I add this to my shard, what types of issues should I be looking for?

Tasanar:
I just meant as I was testing is there anything I should be looking for?

Vorspire:
Yes, there is one test you can do, the BlackthornDungeon region has a 5-second interval timer that moves idle players;
`
        public void OnTick()
        {
            foreach (Mobile m in this.GetEnumeratedMobiles().Where(m => m is PlayerMobile && m.AccessLevel == AccessLevel.Player))
            {
                if (m.Hidden)
                    m.RevealingAction();

                if (m.Y > 2575 && m.LastMoveTime + 120000 < Core.TickCount)
                    MoveLocation(m);
            }
        }
`

See if you can trigger the above code, it had an issue at one point where it would cause a crash, but it should be prevented in this PR.

Tasanar:
Okay will test now and report back.

The overall fix; if you could explain it like you would explain it to a 10 year old,  what does it do?

Vorspire:
Mainly it reduces memory usage (no longer caching entities in a list), and unnecessary CPU overhead (removing parallel tasks).

Tasanar:
`  public void OnTick()
        {
            foreach (Mobile m in this.GetEnumeratedMobiles().Where(m => m is PlayerMobile && m.AccessLevel == AccessLevel.Player))
            {
                if (m.Hidden)
                    m.RevealingAction();

                if (m.Y > 2575 && m.LastMoveTime + 120000 < Core.TickCount)
                    MoveLocation(m);
            }
        }  `

I went into blackthorns dungeon on a test player and left him AFK. He was force moved and it did not crash the server. So everything is fine there.

mem

Right off the bat too it seems to be using less memory. Before on start up it would avg around 740 - 750. This one started up at 725.

PR #3342, opened 2018-04-16
Update Rewards.cs

Tasanar:
The method to prevent murders was commented out.

This leaves open a loop hole where reds could get into Eodon.


Tasanar:
Not that I can think of? Murders can not go to Eodon. This little "exploit" allows them in.

kevin-10:
Haha, wasn't an exploit, just an oversight on my part. On Alexandria, reds can go to any facet, just forgot to remove that point.

PR #3343, opened 2018-04-17
Custom Housing Optimization

Vorspire:
+ Prototyping optimizations for house customization.

### DIFF
--- a/Scripts/Multis/HouseFoundation.cs
+++ b/Scripts/Multis/HouseFoundation.cs
@@ -995,12 +995,20 @@ namespace Server.Multis
 
         public bool IsHiddenToCustomizer(Mobile m, Item item)
         {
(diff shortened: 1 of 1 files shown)


kevin-10:
I debated hiding other houses, it looked too custom to me.  Either way I'm good.

Tasanar:
So no houses will show? That will definitely stick out to EA players. Dexters edits did fix the problem but if this is needed, so be it.

Vorspire:
I don't see a reason why someone customising a house needs to be aware of other houses, and from what I've read on the matter so far, OSI haven't been consistent in fixing the issue. 

There is the added benefit of being able to build without an obstructed view if a large house is placed in front of yours.

Tasanar:
Great point.

Merging now.

PR #3348, opened 2018-04-18
Crash Fix

kevin-10:
- Enabled Spell Instant Target to invoke an instance of the spells target, that way any conditions embedded in the OnTarget method will be checked. This should also fix the MethodInfo crash by complete removing it. Added logging for unsupported target constructor args, in which would be fixed once identified.

@Tasanar, periodically check your server root folder for InstantTargetErr.log. This will identify any spells that use a target, that don't have the standard constructor.


Tasanar:
Sure! I am going to approve it and add it right now. As soon as any logs are created I will report them in Bug Reports.

PR #3356, opened 2018-04-20
Missing Cloaks of Life, Death, and Power

Tasanar:
http://www.uoguide.com/Cloak_Of_Power
http://www.uoguide.com/Cloak_Of_Death
http://www.uoguide.com/Cloak_Of_Life
http://www.uoguide.com/Cloak_Of_Silence

Adding missing cloaks from EA event. All hues verified on EA. Correct clilocs.

This item derives from BaseOuterTorso because although it is an Elven Cloak, humans can wear it too.

### DIFF
--- a/Scripts/Items/Artifacts/Equipment/Clothing/CloakOfDeath.cs
+++ b/Scripts/Items/Artifacts/Equipment/Clothing/CloakOfDeath.cs
@@ -1,0 +1,40 @@
+using System;
+
+namespace Server.Items
+{
+    [Flipable(0x2FB9, 0x3173)]
+    public class CloakOfDeath : BaseOuterTorso
+    {
+        [Constructable]
+        public CloakOfDeath()
+            : base(0x2FB9)
+        {
+            Weight = 2.0;
+			Hue = 0x966;
+			Attributes.DefendChance = 3;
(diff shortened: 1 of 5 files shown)


kevin-10:
Before you merge, make sure you add it to the csproj file.

Tasanar:
How do I do that correctly? I have looked at it before but I never know where to put the correct edits.

Xeroxxx:
If you're using Visual Studio, basiclly drag and drop it to the correct path in the solution explorer.

If using a text editior open Scripts.csproj and add the line:
`<Compile Include="Path\Path\File.cs" />` to an appropriate place. Preferably sorted by alphabet

Tasanar:
Just looking at it I am not sure where to put it. It does not seem to be sorted.

Xeroxxx:
Put it where you want it and it makes sense. Actually doesnt matter.

Xeroxxx:
Maybe just before:
`    <Compile Include="Items\Artifacts\Equipment\Clothing\TheScholarsHalo.cs" />
    <Compile Include="Items\Artifacts\Equipment\Clothing\JadeArmband.cs" />
    <Compile Include="Items\Artifacts\Equipment\Clothing\MagicalDoor.cs" />`

PR #3372, opened 2018-04-24
Update Main.cs Honesty.cs

Xeroxxx:
Cosmetics

+ Updated output of Honesty to be inline with rest of ServUO
+ Calculating BuildDate with Build & Rev to improve Support and roughly findout when that build is from
- Removed useless output when ServUO is started (------)


Xeroxxx:
capture

PR #3377, opened 2018-04-26
Delete MailboxEastAddon.cs

Tasanar:
This is a duplicate of the Clean Up Brit mail box.

I tested and confirmed it does not have the same function and does not interact with the real mail box in any way.


kevin-10:
Please have it removed from teh csproj.

PR #3378, opened 2018-04-27
Leveling system

drew323f:
want to use as a base system


Tasanar:
This will most likely not get pushed to the main repo. ServUO has the goal of emulating real UO, not adding custom content.

Vorspire:
Looks great, but custom features such as this cannot be redistributed with the main repository and may cause conflicts for those who already use custom level systems.

I recommend releasing your work as a resource on http://www.servuo.com

PR #3390, opened 2018-05-06
Bug Fixes

kevin-10:
- Enhanced/Crafted items no longer double resource luck Bonus
- SpawnRange now is serialized on Spawner.cs
- Fixed MagicKey Crash


Argalep:
Luck values ​​of the items have fallen. Is this normal?

PR #3394, opened 2018-05-09
Bunch of misc issues found using PVS-Studio

DarkLotus:
Some random errors, a few logic ones im not sure of ie Stealing, if someone wants to review/edit.

### DIFF
--- a/Scripts/Mobiles/NPCs/MysteriousWisp.cs
+++ b/Scripts/Mobiles/NPCs/MysteriousWisp.cs
@@ -497,7 +497,7 @@ namespace Server.Mobiles
                 {
                     Item item = m_Items[idx];
 
-                    if (m_Wisp.Backpack.Items.Contains(item) && !item.Deleted && item != null)
+                    if (item != null && m_Wisp.Backpack.Items.Contains(item) && !item.Deleted)
                     {
                         from.SendGump(new InternalGump2(m_Wisp, item));
                     }
--- a/Scripts/Services/Expansions/Time Of Legends/Mobiles/DragonTurtle.cs
+++ b/Scripts/Services/Expansions/Time Of Legends/Mobiles/DragonTurtle.cs
@@ -174,7 +174,7 @@ namespace Server.Mobiles
                             if (d == Direction.Up)
                                 d = Direction.North;
                             else
-                                d = d++;
+                                d = ++d;
 
                             Movement.Movement.Offset(d, ref x, ref y);
                         }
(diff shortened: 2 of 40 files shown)


Scripts/Services/FireCasino/DiceGames.cs
```diff
@@ -126,12 +126,6 @@ public virtual void Reset()
             BettingOn = -1;
             Stage = GameStage.Betting;
 
-            if (Roll != null)
```
kevin-10: Is this going to cause an issue?

Vorspire: It's fine, Roll is always nullified before being checked for not-null, meaning the if-statement was not necessary.

Scripts/Services/Monster Stealing/Core/StealingHandler.cs
```diff
@@ -166,7 +166,7 @@ private static int GetStealingChance(Mobile thief, BaseCreature from, double ste
 
             if (level >= 40)
                 chance += 5;
-            else if (level >= 40)
+             else if (level >= 40) /*??*/
```
Vorspire: Assume 40, 35, 30.

Scripts/Services/Expansions/Time Of Legends/Mobiles/DragonTurtle.cs
```diff
@@ -174,7 +174,7 @@ public void DoBubble()
                             if (d == Direction.Up)
                                 d = Direction.North;
                             else
-                                d = d++;
+                                d = ++d;
```
Vorspire: Neither the old nor the new lines make sense here, left-hand assignment isn't necessary.

Scripts/Mobiles/NPCs/MysteriousWisp.cs
```diff
@@ -497,7 +497,7 @@ public override void OnResponse(NetState state, RelayInfo info)
                 {
                     Item item = m_Items[idx];
 
-                    if (m_Wisp.Backpack.Items.Contains(item) && !item.Deleted && item != null)
+                    if (item != null && m_Wisp.Backpack.Items.Contains(item) && !item.Deleted)
```
Vorspire: Wisp Backpack should be null checked too.

Scripts/Services/FireCasino/DiceGames.cs
kevin-10: Haha, I should know, I wrote that!  It's been a while though.  Good catch.

DarkLotus: Once i realized how many of these unneeded null checks there were(100+) i stopped removing them, none are in hot paths so shouldn't be a performance issue leaving them.

Scripts/Services/Monster Stealing/Core/StealingHandler.cs
DarkLotus: Will fix, i did browse uoguide and couldn't find anything related to fame for the formula so i was unsure of this one.

PR #3433, opened 2018-05-20
Invisibility, Darkglow, and Parasitic Poison potion can be added to p…

dohertki:
…otion kegs


dohertki:
ok

kevin-10:
Also, I have not seen any darkglow/parasitic potion kegs on EA. Do they exist there?

Tasanar:
No they 100% do not. I just searched on Atlantic. They are all there on commodity deeds but do not appear on kegs. This should not get pushed.

PR #3473, opened 2018-06-01
Update LowlandBoura.cs

Tasanar:
Left over code from a creature that does not even drop the item.


kevin-10:
1/1000 chance on those?

Tasanar:
oh, was it already 1 in 100?

Tasanar:
I posted a link on ServUO. Seemed to be quite rare. They are all VERY easy to kill.

PR #3484, opened 2018-06-02
Bug fix.

Argalep:
* Throw weapons fix.

https://uo.com/wiki/ultima-online-wiki/items/quest-items-rewards/

screenhunter 86

### DIFF
--- a/Scripts/Items/Artifacts/Equipment/Weapons/AbyssReaver.cs
+++ b/Scripts/Items/Artifacts/Equipment/Weapons/AbyssReaver.cs
@@ -5,41 +5,27 @@ namespace Server.Items
     public class AbyssReaver : BaseThrown
 	{
 		public override bool IsArtifact { get { return true; } }
+        public override int LabelNumber { get { return 1112694; } } // Abyss Reaver
+
         [Constructable]
         public AbyssReaver()
             : base(0x901)
         {
-            this.Weight = 6.0;
-            this.Layer = Layer.OneHanded;
-            this.Hue = 1195;
+            Weight = 6.0;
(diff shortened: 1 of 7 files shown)


Argalep:
If the class is changed, existing items in the world are deleted.

Tasanar:
Well the way it is coded now is terrible. 

I say go for it. It should not have been like that to begin with.

kevin-10:
We could keep it, as long as all the weapon properties match what they are supposed to be.

Tasanar:
But then say UO changes a property of the Boomerang or thrown items down the road, just one more side edit that would need to happen.

Tasanar:
This not the only item like this from SA. Is just re-writing each class for each of these worth it? Let existing shards figure out their own way to replace it while cleaning it up for newer users as much as possible.

Tasanar:
I will dismiss my review but I think we should have just made it right, here and now, instead of applying a bandaid fix. Who cares if this one item got deleted. Owners could set up their own way of re-adding it to the correct players after it was corrected.

Vorspire:
Make it inherit from the correct class hierarchy. 

There are ways to do hierarchy insertion without wiping all objects of that type - take a look at how Power Scrolls, Alacrity Scrolls and Transcendence Scrolls had SpecialScroll inserted into their hierarchy.

kevin-10:
Yea, I was just thinking that. I had to do that to a bunch of items for the new bod system. @Argalep, can your just cut out the AbyssReaver stuff and I can get a fix in. @Tasanar, can you let me know which items will need this for?  Thanks!

PR #3506, opened 2018-06-08
Update CitadelTeleporters.cs

Tasanar:


Scripts/Items/Internal/CitadelTeleporters.cs
```diff
@@ -50,6 +36,7 @@ public override void OnDoubleClick(Mobile from)
                 PlayerMobile player = (PlayerMobile)from;
 				
                 if (QuestHelper.GetQuest(player, typeof(BlackOrderBadgesQuest)) != null || QuestHelper.GetQuest(player, typeof(EvidenceQuest)) != null)
+                    BaseCreature.TeleportPets(player, new Point3D(107, 1883, 0), Map.Malas);
```
kevin-10: You need brackets here.

Scripts/Items/Internal/CitadelTeleporters.cs
```diff
@@ -10,28 +10,14 @@ public class CitadelTele : Item
...
         {
-            this.Movable = false;
+            Movable = false;
         }
 
         public CitadelTele(Serial serial)
             : base(serial)
         {
         }
-
```
kevin-10: Why are we removing decay?  Don't these decay and re-spawn on EA?

Tasanar:
Why do brackets need to be there? They weren't there originally. 

It also works fine and compiled fine on my server.

Please see the bug report I posted on the forums about the crate. There is no information saying it decays. Also why would one of the entrances randomly decay?

kevin-10:
If you have one line that executes from an if statement, you don't need brackets.  If you have more than 1, you have to have brackets or it won't execute properly.

kevin-10:
Once it decays, the spawner spawns a new one.  I remember it moving locations on EA.  I would have coded  it where teleport around on its own without deleting. I don't like relying on spawners for this stuff.

Tasanar:
I'll add the brackets, guess it never worked right.

I posted a few links on the crate over at ServUO.

kevin-10:
YOu only need them when you added the new code. Before it was only 1 line, so it was ok.

kevin-10:
They were added on the 2nd one. We're good.  How about the crate moving?  Should we investigate that on EA?  We can remove the decay, and if we verify the crate moves, we can have it actually instead of deleting/decaying and respawning.

Tasanar:
Yes of course. We can continue this over on the Bug Report forums at ServUO. Had questions there about how there coded in regards to followers anyways.

PR #3589, opened 2018-06-26
Bug fixes

kevin-10:
- Fixed object properties on mining cart and sheep statue
- Fixed erroneous gump on Runic Atlas
- Added Cusidhe to Wolf slayer
- Seed Box should always wight 10 stones
- Necro AI will no longer cast corpse skin in that creature has physical or cold damage Type
- Arena Manager should no longer offer animal training Quest
- Added skill cap to animal lore Gump
- Fixed navreys pillar during deserialization
- Davies lovker now shows contents in property list
- Vile Water Elementals can no longer move over canals

Enhancements:
Consolidated several functions in BaseCreature, using CreatureKilled/KilledBy events to clean up code. Basically, imbuing drops, ML arty drops and TOL recipe/armor set drops are now handed in thier respected systems.

### DIFF
--- a/Scripts/Items/Addons/MiningCart.cs
+++ b/Scripts/Items/Addons/MiningCart.cs
@@ -16,14 +16,16 @@ namespace Server.Items
 
     public class MiningCart : BaseAddon, IRewardItem
     {
+        public override bool ForceShowProperties { get { return true; } }
+
         public override BaseAddonDeed Deed
         {
             get
             {
                 MiningCartDeed deed = new MiningCartDeed();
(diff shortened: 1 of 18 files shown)


Tasanar:
If you use OnKilledBy, it is invoked for each player that did damage and has loot rights, but CreatureDeath is invoked only once.

That is what Vorspire said.

So the problem I see now is creatures could award multiple artifacts on death, to multiple people, when there should only be ONE chance.

kevin-10:
Yes, I know that.  OnKilledBy is used for dropping ML mini arties, where before this update, happen to happen in the same function OnKilledBy was invoked.

PR #3592, opened 2018-06-26
Update BlanketOfDarknessArtifact.cs

Tasanar:


kevin-10:
I would recommend having it derive from BaseDecorationArtifact. If its too laste, you can always mimic its functionality.

kevin-10:
Approved so we're not hanging folks compiling up.

PR #3604, opened 2018-06-29
Tasanar patch 2

Tasanar:
All Armor and Jewelry properties should now be in the correct order when compared to EA.

Will be working on Clothing and weapons next. Weapons will take some time as they have a slightly different order, for whatever reason.


kevin-10:
Yea, and for weapons, be careful of the property mods added by certain spells, etc.

Tasanar:
Yeah I will. I'll work on weapons slowly and will probably do it in a few pushes.

PR #3611, opened 2018-07-02
Peerless Keys are blessed on EA

Tasanar:
http://www.uoguide.com/Sabrix%27s_Eye
https://trueuo.com/index.php?threads/citadel-peerless-keys-arent-blessed.654/

Also confirmed on EA. Working on confirming the other peerless.


kevin-10:
Aren't all peerless keys blessed?  If so, we can just mark it in PeerlessKey.

Tasanar:
I can not confirm the crystals. Also the Exodus keys are not blessed and not sure about Medussa or Stygian dragon. They all derive from PeerlessKey

PR #3639, opened 2018-07-07
Update DisintegratingThesisNotes.cs

Tasanar:
Need to revert this one key being blessed.

The problem is on EA they ARE blessed but just appear on the corpse. The way we have loot drops, blessed items will not drop on the corpse this way.

Reverting until a better system is discovered.


Tasanar:
Tested and it compiled fine! Thanks Vorspire.

Issue #3642, opened 2018-07-08
AutoSave.cs deletes the /Saves folder on Linux


Elyrith:
System:
Ubuntu 18.04
mono-complete 4.6.2 (Debian 4.6.2.7+dfsg-1ubuntu1)
Followed some of the stuff from https://www.servuo.com/tutorials/linux-mono.40/ but not much.

Issue:
Whenever the server does an autosave it does a Directory.Move and moves the /Saves folder into the appropriate Backups folder, but now the /Saves folder is gone. Even if you run a [save (or "save" in console) it doesn't re-create it.

This means if you shutdown or restart, when you start it back up it acts like a completely new server again (asking to create a new admin user).

PR #3643, opened 2018-07-08
Add a recursive copy function to the autobackup

Elyrith:
Add a recursive copy function to the autobackup, rather than moving the folder.

I tested this on my Linux (Ubuntu 18.04) server and my Windows 10 server and the backups worked without losing my save data. (I was having no issues on Windows, but I tested it there to make sure I didn't break it.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,8 @@ PHONY : default build clean run
 default: run
 
 debug: 
-	${MCS} -target:library -out:${CURPATH}/Ultima.dll -r:${REFS} -nowarn:${NOWARNS} -d:DEBUG -d:ServUO -d:NEWTIMERS -nologo -debug -unsafe -recurse:${SDKPATH}/*.cs
-	${MCS} -win32icon:${SRVPATH}/servuo.ico -r:${CURPATH}/Ultima.dll,${REFS} -nowarn:${NOWARNS} -target:exe -out:${CURPATH}/ServUO.exe -d:DEBUG -d:ServUO -d:NEWTIMERS -nologo -debug -unsafe -recurse:${SRVPATH}/*.cs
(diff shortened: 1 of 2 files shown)


Elyrith:
This *should* fix issue #3642

Elyrith:
For the record, I 'borrowed' the code from here.
https://docs.microsoft.com/en-us/dotnet/standard/io/how-to-copy-directories

Vorspire:
I've made the necessary edits, please test on MONO again.

Elyrith:
It was still using the non-Mono version before I added the flag for the make script.

I got the info from:
https://stackoverflow.com/questions/4406211/compiler-if-directive-to-split-between-mono-and-net

Elyrith:
Wait, still testing...

Elyrith:
This works for me in Linux with Mono.

I got my info from:
https://stackoverflow.com/questions/4406211/compiler-if-directive-to-split-between-mono-and-net

Vorspire:
Everything in the ServUO project references the MONO directive, it would be best to stick with the status quo and use that instead of __MonoCS__ 

The make file included with ServUO should already define #MONO, as does the compile batch file for compiling for MONO on Windows.

The edits I made should have been enough, should we be updating the rest of ServUO to use __MonoCS__ instead? @Argalep What do you think?

Argalep:
I searched a bit for the eye. it seems to be a better way to use _MonoCS_ instead of mono in the project.

Elyrith:
Perhaps it's a bad idea after all, since it looks like __MonoCS__ isn't supported in Mono 5.0.0 anymore.
https://www.mono-project.com/docs/about-mono/releases/5.0.0/

In a moment I'll make an update that uses the check from the Mono documentation here:
https://www.mono-project.com/docs/faq/technical/#how-can-i-detect-if-am-running-in-mono

PR #3720, opened 2018-07-20
Bug fixes

kevin-10:
- Special Scroll Book contents no longer cound towards item count or weight
- Tinker tool small bods no longer come with resource requirement
- Mining gloves can now be reforged properly
- Fixed issue (hopefull) where under certain circumstances, more than 3 cached bods would be given to a Player
- Travesty will not attack with a ranged weapon if it clones a player with a ranged Weapon
- Travesty now uses body/hue mod

Fixed Travesty Crash
- Travesty now uses music skills of morphed player


Tasanar:
Woops not sure why that happened.

New Haven - Only the undead do not attack. Everything else on new haven island, like the brigands, ect, will attack if not Young.

PR #3742, opened 2018-07-28
Siege Mode moongate.cs update

LordFenris:
The file did not contain the Siege Shard functionality verification , as a consequence of which the PK characters did not have the ability to move through the non-public moongate outside of Felucca.


kevin-10:
Thanks!

PR #3759, opened 2018-08-01
Housekeeping

Vorspire:


Scripts/Items/Books/SpecialScrollBooks/BaseSpecialScrollBook.cs
```diff
@@ -1,182 +1,214 @@
...
+        {
+            base.Serialize(writer);
+
+            writer.Write(2); // version
+
+            writer.Write((int)Level);
+            writer.Write(_Capacity);
+        }
+
+        public override void Deserialize(GenericReader reader)
```
Tasanar: https://github.com/ServUO/ServUO/pull/3759/files#diff-075c2fb9a54d1f04acb1b5d48220a638

Something here is preventing shards with existing powerscroll books, or probably any special scroll book, for loading up without deleting them. Otherwise everything else looks great!

PR #3813, opened 2018-08-23
Update EngravingTools.cs

Tasanar:
Fished up MetalGoldenChest could not be engraved. They can be on EA


kevin-10:
What about htat other container you can fish up, the rare one?  Lockbox i think it is?

Tasanar:
I have no idea on that one. I'll look around on Atlantic and see if I can find any ingraved in a house.

PR #3818, opened 2018-08-26
* CandlewoodTorch fix

Argalep:


Tasanar:
Argalep myself and Dexter have been using my forums until ServUO is up.


https://trueuo.com/index.php?forums/servuo-bug-reports.75/

PR #3839, opened 2018-08-31
Crash Fix + other fixes

kevin-10:
- One more attempt at Mini Champ System bug where champ doesnt spawn and the system gets hung up
- Removed redundant OnEnter/OnExit methods in Region/Mobile:

@Vorspire
Here is the problem with the Region issue.  I can see why the redundant checks were added, so it would be called at the appropriate timing, ie OnExit called before the new region was set to the mobile, and OnEnter begin called aferwards.  The problem is, in Region.cs, it is needed there to be called during the appropriate child/parent regions as well.  This will need to be looked into in the near future.

### DIFF
--- a/Scripts/Services/MiniChampionSystem/MiniChampionSpawn.cs
+++ b/Scripts/Services/MiniChampionSystem/MiniChampionSpawn.cs
@@ -21,7 +21,10 @@ namespace Server.Engines.MiniChamps
         [Description("MiniChampion Generator")]
         public static void GenStoneRuins_OnCommand(CommandEventArgs e)
         {
-            Controllers.ForEach(x => x.Delete());
+            foreach (var controller in Controllers)
+            {
+                controller.Delete();
+            }
 
             Map map = Map.TerMur;
 
@@ -200,8 +203,7 @@ namespace Server.Engines.MiniChamps
             m_Level = 0;
 
             ClearSpawn();
-            Despawns.ForEach(x => x.Delete());
-            Despawns.Clear();
+            Despawn();
 
             if (m_Timer != null)
                 m_Timer.Stop();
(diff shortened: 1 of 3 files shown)


Argalep:
`Exception:
System.NullReferenceException: Object reference not set to an instance of an object.
   at Server.Spells.SkillMasteries.FistsOfFurySpell.OnDamaged(Mobile attacker, Mobile defender, DamageType type, Int32& damage) in ..\Scripts\Spells\Skill Masteries\FistsOfFury.cs:line 91
   at Server.Spells.SkillMasteries.SkillMasterySpell.OnDamage(Mobile victim, Mobile damager, DamageType type, Int32& damage) in ..\Scripts\Spells\Skill Masteries\Core\SkillMasterySpell.cs:line 749
   at Server.AOS.Damage(IDamageable damageable, Mobile from, Int32 damage, Boolean ignoreArmor, Int32 phys, Int32 fire, Int32 cold, Int32 pois, Int32 nrgy, Int32 chaos, Int32 direct, Boolean keepAlive, DamageType type) in ..\Scripts\Misc\AOS.cs:line 336
   at Server.AOS.Damage(IDamageable m, Mobile from, Int32 damage, Int32 phys, Int32 fire, Int32 cold, Int32 pois, Int32 nrgy, Int32 chaos, Int32 direct, DamageType type) in ..\Scripts\Misc\AOS.cs:line 97
   at Server.Spells.SpellHelper.SpellDamageTimerAOS.OnTick() in ..\Scripts\Spells\Base\SpellHelper.cs:line 1533
   at Server.Timer.Slice() in ..\Server\Timer.cs:line 386
   at Server.Core.Main(String[] args) in ..\Server\Main.cs:line 673`

PR #3868, opened 2018-09-09
Khaldun

Argalep:
see (https://www.uo-cah.com/pub-101/khaldun-champion-spawn)


kevin-10:
Is there a setup for all this?

Tasanar:
on real UO they announced it would be a seasonal thing. I would caution against an automated set up yet. It is also still on Test Center.

PR #3873, opened 2018-09-10
Tasanar patch 1

Tasanar:
Adds the missing system messages, corrects a few others, updates range from 1 to 2 to both open and use the gump. Tested on EA.


Tasanar:
All credit goes to Lokai for the boat edit.

Under the old decay system it NEVER said like new because it was always >= Also this way allows shard owners to easily set decay times and not have to edit this method either. Tested with timers of 2 minutes and 10 minutes and it works great.

PR #3889, opened 2018-09-13
Update HouseSign.cs

Tasanar:
When players change their name and perform some actions sometimes there is a delay in the house sign updating. This simply updates the sign when someone double clicks on it and it passes all the checks.

### DIFF
--- a/Scripts/Multis/HouseSign.cs
+++ b/Scripts/Multis/HouseSign.cs
@@ -202,6 +202,7 @@ namespace Server.Multis
                 }
             }
 
+            InvalidateProperties();
             ShowSign(from);
         }
 


kevin-10:
You mean the house sign as in the mouse over of the actual house sign item?

Tasanar:
Yeah, like change your name using a name change and sometimes other props lag. I use this and no issues now.

kevin-10:
Yea, i mean it won't hurt anything, but its probably not going to solve the problem in some cases. What if a player simply mouses over the sign after the owner changes his name w/out double clicking it?  Maybe a line needs to be added in PlayerMobile after a name change?

kevin-10:
foreach(var house in BaseHouse.Houses.Where(h => h.Owner == this && h.HouseSign != null))
{
    house.HouseSign.InvalidateProperties();
}

Something to that effect.

Tasanar:
Mouse over won't update. This way if someone just clicks on the sign it auto refreshes any outstanding properties. Just a simply fix.

PR #3904, opened 2018-09-16
* add CounterfeitPlatinum

Argalep:
* add khaldun chest
* CultistsRitualTome hue fix.
* add PolishedMeteorite

### DIFF
--- a/Scripts/Items/Resource/PolishedMeteorite.cs
+++ b/Scripts/Items/Resource/PolishedMeteorite.cs
@@ -1,0 +1,33 @@
+using System;
+
+namespace Server.Items
+{
+    public class PolishedMeteorite : Item
+    {
+        public override int LabelNumber { get { return 1158693; } } // polished meteorite
+        
+        [Constructable]
+        public PolishedMeteorite()
+            : base(41422 + Utility.Random(12))
+        {
+            Weight = 3;
+        }
+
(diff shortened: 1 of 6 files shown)


Argalep:
there are missing parts:

* Caddellite Crafting
* Caddellite Infused

i do not know them.

kevin-10:
A couple things.

1. Why is the chest's ID flags set to none?  This may create some issues since it wont be flagged as a container anymore, especially since other containers use this ID.

2. You don't need a region to created that message. That message is generated in DetectHidden if the chest inherits IRevealableItem.

Argalep:
Players must walk through the secret crates. how do we do it.

Argalep:
@kevin-10 

If we put a hidden item in which the players can pass, if we put the chest in place when the secrecy of the item is removed, will it solve our problem?

kevin-10:
I'm not sure. I'm looking at ItemData now, to see if we can override it for individual items, instead of item id's. We can run it as is, and address this next.

kevin-10:
Are these items only passable when hidden in EA?

Argalep:
Yes

> Are these items only passable when hidden in EA?

yes

kevin-10:
So, if they are not hidden, they are not passable, like a decorative box?

Argalep:
Yes

PR #3911, opened 2018-09-19
Basket Of Green Tea Fix

Tasanar:
As of Publish 96: http://www.uoguide.com/Publish_96

Basket of Green Tea now weigh 1 stone and will stack. This only applies to newly purchased Baskets, old baskets will not stack with new Baskets.
Innkeepers in Zento will now stock 1000 Baskets of Green tea.
Increased the number of Innkeepers that spawn in the Zento Inn to 3.


Tasanar:
Slasher has not dropped Breastplate in quite some time.

Confirmed on EA.
http://www.uoguide.com/Breastplate_of_the_Berserker

PR #3918, opened 2018-09-21
Abyss Reaver

Tasanar:
When awarded as a quest reward it did not appear because it could not actually be constructed in-game. This fixes that.


kevin-10:
Damn SA code!  There were a bunch of these, I thought we got them all.  Thanks!

PR #3958, opened 2018-10-04
Gargoyle Necklaces and Earrings

Tasanar:
These items should NOT be able to be enhanced. Confirmed on EA. Although they take up jewelry slots they already have resists and allow gargoyles to really go over cap in resists, compared to any other race. Enhancing just furthers that advantage.

https://stratics.com/threads/enhancing-gargoyle-jewelry.267417/
https://community.stratics.com/threads/wasted-a-forged-metal-tool-charge-on-a-gargoyle-amulet.328672/

### DIFF
--- a/Scripts/Services/Craft/Core/Enhance.cs
+++ b/Scripts/Services/Craft/Core/Enhance.cs
@@ -53,6 +53,9 @@ namespace Server.Engines.Craft
         {
             if (item == null)
                 return EnhanceResult.BadItem;
+			
+			if (item is GargishNecklace || item is GargishEarrings)
+                return EnhanceResult.BadItem;
 
             if (!item.IsChildOf(from.Backpack))
                 return EnhanceResult.NotInBackpack;


Tasanar:
Because these are not derived from BaseJewelry and instead BaseArmor this was always overlooked.

kevin-10:
First thing, they don't take jewelry slots, they take gloves/neck slot, so any suggestion of being over powered is the same as saying enhancing a gorget or gloves being over powered when enhanced. Also, are we sure, as in, have we recently checked, they can't be enhanced? I honestly don't see why they wouldn't be able to be enhanced.

kevin-10:
This would be one of those things where I wouldn't mind NOT following EA on. Like i said, this gives no benefit to gargoyles by letting these items be enhanced since they use standard armor slots.

kevin-10:
It actually nerfs gargoyles, and as you can see by those posts, many players are pretty upset that they won't address this issue.

kevin-10:
Thoughts?

Tasanar:
Well I do agree with doing things BETTER than EA, that being said I thought they had an advantage with resists as it stands especially with Stone armor?

kevin-10:
OK, I didn't know that. Looks like a 15 resist difference between stone armor and plate armor.  Wow. Fucking EA man. To make it even more jacked up, for this justification (which im assuming why EA wont let these be enhanced), its assuming that the person is wearing stone armor, which isn't always true!

kevin-10:
We'll need to remove resource application when crafted, as well.

Tasanar:
Maybe stone should be an armor type just like plate is?

PR #3967, opened 2018-10-05
new magincia bazaar gump fix.

Argalep:


Argalep:
If you take the rustic package to the osi account, I will do the same :)

Tasanar:
What do you mean?

Argalep:
Need to have a rustic or gothic package to do more testing in osi

Argalep:
image

PR #3968, opened 2018-10-05
* add ICommodity item

Argalep:


kevin-10:
Don't all the reagents have ICommodity individually?

Argalep:
yes, does this cause problems?

Tasanar:
If they all use it, no, but the redundant check in each one should then be removed I am guessing.

kevin-10:
Or just remove the one in BaseReagent.

PR #3971, opened 2018-10-06
Bug Fixes

kevin-10:
- Armor Refinement Max Resist now trumps stone form max resist, entirely
- Armor Refinement Max Resist taken into account when cursed
- Added 3 second cooldown to fly spell (2 seconds to cast, 1 second actual cooldown)
- Adding seeds to seed box will auto invalidate original seed properties for display purposes
- Hidden containers no longer block movement


Argalep:
Have the following code in the chests to remove it?

`public static void Initialize()
        {
            TileData.ItemTable[0x2DF3].Flags = TileFlag.None;
        }`

kevin-10:
No, that would make them passable at all times, and they wouldn't be marked as containers anymore.

PR #4002, opened 2018-10-17
Shadowguard Fixs.

Argalep:
* When the boss spawns continue spawning the Liquor bottles
* When all the trees are destroyed, spawns is deleted.
* Canals are completed, spawns is deleted.
* Phylactery, corpse does not fall

reference : https://www.youtube.com/watch?v=57TnLVVRZco


kevin-10:
Instead of making _Completed public, just make a public property:

public bool Completed { get { return _Completed; } }

kevin-10:
Otherwise, well done. I'm assuming this is just based on EA accuracy.

PR #4024, opened 2018-10-27
Artifact Armor, Clothing, and Jewelry

Tasanar:
1. Now have all the correct Clilocs.
2. Fixed some wrong stats.
3. Removed all the this. statements.


Tasanar:
Adds IsArtifact tag to a few ones that were missing.

PR #4031, opened 2018-10-27
fixs.

Argalep:
* Travesty Spawn Helper Fix.
* Gargoyle items are not loot

### DIFF
--- a/Scripts/Mobiles/Bosses/Travesty.cs
+++ b/Scripts/Mobiles/Bosses/Travesty.cs
@@ -382,16 +382,11 @@ namespace Server.Mobiles
             }
         }
 
-        public static void SpawnNinjaGroup(Point3D _location)
+        public void SpawnNinjaGroup(Point3D _location)
         {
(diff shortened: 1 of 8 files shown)


Tasanar:
The specific way Hephaestus is created right now, Antique vs not, will also make altering it challenging. Just one example.

Argalep:
Hephaestus may not be an antique item.

screenhunter 173
screenhunter 174
screenhunter 175

Tasanar:
If it comes from the Void pool it is antique.

If it comes from Cora it is not.

Also there are a bunch of TOL items currently that do not alter correctly if we approve this.

Tasanar:
So lastly what about the Hephaestus shield?

If you alter an Antique one will it still have Antique and vice versa?

Argalep:
Items received as a reward cannot be alterabled. That is, you cannot alter the antique.

Tasanar:
But how does it know? They are both derived from the same item

PR #4043, opened 2018-10-30
Bug Fixes

kevin-10:
- Removed gold/checks as refunds from player/rented vendors
- Gargoyle penalty with throwing weapon/shield is now reduced from overall change to hit, as opposed to the hit chance increase. Per EA, this will significantly reduce the chance to hit with a shied without parry skill.


kevin-10:
Should what be 90 out of 120?

PR #4046, opened 2018-10-31
Costume fixes

Tasanar:
1. Wolf spider costume had an invalid bodyvalue resulting in players appearing to be "invisible".

2. BaseCostume should not be IDyable.


Tasanar:
https://uo.com/wiki/ultima-online-wiki/items/slayer-property-items/

"Harpy and Stone Harpy are not birds, they have no slayer vulnerability"

PR #4048, opened 2018-10-31
* New Magincia Marketplace Text fix.

Argalep:


kevin-10:
Is this tested with all the different possible messages? I have a workaround that works. Also, how did you know to put "empty"?  Packet sniffers?

Argalep:
The client is having problems when the parameter is null. I saw that it works correctly in any word. If you have a better solution, we can try it.

kevin-10:
For the one with an empty string, I used the \t and it worked.

kevin-10:
How did you know to use the @ symbol?

Argalep:
I have encountered the problem found in the Ultima Store menu.  OSI is also used this way


https://github.com/ServUO/ServUO/blob/master/Scripts/Services/UltimaStore/UltimaStoreGump.cs

PR #4133, opened 2018-12-04
* Initial inclusion of .net 4.6.1, c#6 and roslyn script compiler

Argalep:


kevin-10:
@Vorspire im gonna let you handle this one lol.

ggobbe:
Any news regarding this MR? @Vorspire @kevin-10

Xeroxxx:
We removed the usage of #if MONO statements. This won't work as intended.

Tasanar:
@Vorspire is this good to go or should I close it out. It has been a few months now.

dmurphy22:
I have used `DotNetCompilerPlatform` at work and have had zero problems with it. I think we should go ahead with this. The mono switch is a nice touch as I have no experience with it on mono.

Server/ScriptCompiler.cs
```diff
@@ -210,9 +210,13 @@ public static bool CompileCSScripts(bool debug, bool cache, out Assembly assembl
 
 			DeleteFiles("Scripts.CS*.dll");
 
-			using (var provider = new CSharpCodeProvider())
-			{
-				var path = GetUnusedPath("Scripts.CS");
+#if !MONO
```
dmurphy22: Has this actually been tested on mono? We don't want to start adding code that can only be compiled by Roslyn to scripts which in turn will break mono servers.

ServUO.exe.config
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <configuration>
   <startup>
-    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0" />
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" />
```
dmurphy22: Should we consider 4.7.2 as it is the latest?

dmurphy22:
Merging into SerUO/Roslyn for testing

dmurphy22:
Tested on Ubuntu 18.04 LTS and scripts compile when using roslyn features using `CSharpCodeProvider`. As @ggobbe mentioned, they support c# 6 features in mono using `CSharpCodeProvider`

I am almost ready to merge this into master. I just want to test on ubuntu 16 and ubuntu 14. Then I want to test .net 4.7.2 for viability.

dmurphy22:
Had no issues with compilation on ubuntu 16 and 14. I consider .net 4.6.1 to be stable as far as compilation goes on all targeted platforms. 

Testing 4.7.2 today.

Tasanar:
Is the plan to eventually merge this into the ServUO master?

Tasanar:
Nm scrolled above.

Does it enhance performance in any way? I imagine it will quite the upgrade.

dmurphy22:
Performance should not be affected as the current codebase will not yet implement features found in the later versions of .net. There is the opportunity however to eventually implement these features to improve performance. Networking has seen improvement in .net 4.7 which would be nice to look into.

Compiling scripts.dll is faster using the new compiler. 

The upgrade _should_ be seamless once I merge probably later today.

Server/ScriptCompiler.cs
dmurphy22: Tested and works.

dmurphy22:
Tests went well on 4.7.2

This is ready for further community testing. Merging into master.

PR #4179, opened 2018-12-31
Add New Items and Bug Fixs.

Argalep:
* add SantaStatue
* add Deed for a Roasting Pig on a Spit
* add PulledPorkPlatter, PulledPorkSandwich
* add AddonOptionGump (@kevin-10  The deeds that work with RewardOptionGump can be converted to AddonOptionGump. )
*  HolidaysSign Flipable fix.
*  HolidayWreath Flipable fix.
* squared door mat deed fix.


kevin-10:
Why don't we simply use RewardOptionGump then?

Argalep:
> Why don't we simply use RewardOptionGump then?

Because the menus are different

kevin-10:
It would probably be a better idea to just make an AddonOptionGump that derives from RewardOptionGump and keep IRewardOption stuff. There is no need to make IAddonOption when it does the exact same thing as IRewardOption.

PR #4192, opened 2019-01-07
* Crash Fix.

Argalep:


kevin-10:
Won't that throw a compile error?  Hey, I am consolidating this gump under one gump, instead of 3 different gumps.

kevin-10:
I have a PR that handles this.

PR #4195, opened 2019-01-07
* jewelry box fix (it occupies only one house lockdown, regardless of…

Argalep:
… the quantity of items stored within)


kevin-10:
All you need to do is add JeweleryBox to _NoItemCountTable.

PR #4241, opened 2019-02-15
The candals never should have spawned in there to begin with.

Tasanar:
The candals never should have spawned in there to begin with.


Tasanar:
Regular dungeon chests in general are way off right now. Will need to gather some more information.

PR #4265, opened 2019-03-01
Bug Fixes

kevin-10:
- Updated Swarm effects and Duration
- Players can no longer customize/demolish or set houses to private that have active auctions going on
- Added error checking to Mana Regen Rates
- Fixed bug where improper points were showing for pet training

### DIFF
--- a/Scripts/Services/Pet Training/Gumps.cs
+++ b/Scripts/Services/Pet Training/Gumps.cs
@@ -1365,6 +1365,8 @@ namespace Server.Mobiles
                 }
             }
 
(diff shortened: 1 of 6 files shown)


Scripts/Services/Pet Training/Gumps.cs
```diff
@@ -1365,6 +1365,8 @@ public override void AddGumpLayout()
                 }
             }
 
+            Console.WriteLine("Adjust Value: {0}; Adjust StartValue: {0}", Value, StartValue);
```
Argalep: unnecessary code ?

PR #4268, opened 2019-03-01
Add Triton and Fixs.

Argalep:
* add Triton
* auction safe fix
* NewMaginciaMessageGump fix.
* Vendor Search Map fix

### DIFF
--- a/Config/Honesty.cfg
+++ b/Config/Honesty.cfg
@@ -1,5 +1,5 @@
 # If true, enables virtue honesty
-Enabled=True
+Enabled=False
 
 # Max honesty items that will spawn
 MaxGeneration=1000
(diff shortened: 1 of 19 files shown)


kevin-10:
Approved but I don't like changing BaseGump back go Gump, as BaseGump was supposed to be an improvement over Gump. Not a big deal though.

kevin-10:
Also for the Triton, we should just change IMountStatuette to ICreatureStatuette. I don't see a reason it needs to differentiate between BaseMount and BaseCreature.

Tasanar:
So why was BaseGump changed back to Gump?

kevin-10:
It's not a big deal, but base gump has alot of convenience features and Refresh() which makes is so you can resend the same gump without having to construct a new one.

Tasanar:
I will let @Argalep decide if he wants to push as is, or change it.

kevin-10:
ok cool.  And @Argalep not a big deal, if none of the BaseGump functionality isn't used, we're good.

PR #4272, opened 2019-03-02
UltimaLive addon and specific mods

uoitalia:
This adds UltimaLive (Live map streamer) to the ServUO *CORE* files and scripts, it's not a simple change.

It has new methods to make it work correctly with the hue of the statics, plus adds many more things, usable for static/land searches and work (IPooledEnumerable on MAP)

Also adds the ability to ADD a static/land area, thanks to the use of a specific and custom CommandImplementor and ObjectHandler (plus an utility to allow handling various numbers on the same command implementor), it also adds some functions to the TileMatrix, making it possible to share same map on virtual maps (there are samples written there), and those mods are also to change files in masse live.

The ClientFileExport command is also better, but also SLOWER than in the original ultimalive, it checks every and all statics and land, performs a various cycle, to CLEAN the map and statics (duplicate, invalid, etc) by using the ultima.dll and much more.

The LiveFreeze, with the area command implementor, will allow to freeze an entire zone, but you should be careful when using it, even if I added every possible check, maybe you should make sure BEFORE using it, that it suits your needs as it is now!

I of course consider this mod maybe *too* extreme for the ServUO base, but I'm putting here this 'pull request' to allow everyone that want something like this, to add it on their files.


Tasanar:
Yeah this probably won't get pushed, maybe make a separate branch for it. There is no EC support and the main focus of ServUO is to mostly provide the base UO expansions / experience.

dmurphy22:
Thanks for your contribution @uoitalia but @Tasanar is correct, this is not on mission for ServUO. We also do not want to end up maintaining more third party code. 

The suggestion to fork is a good one or you can submit your changes to the ServUO.com archive.

PR #4302, opened 2019-03-13
Fixes a MAJOR issue with rune books. 

Tasanar:
Bug Fixes

1. This PR addresses a major issue with Runebooks and the Runic Atlas where ANYONE with access to a locked down runebook could drop all the runes and rename the book.

Public runebook libraries have the access set to Everyone, so everyone can use them (obviously) there was NEVER a check to see if the item was locked down. This fixes that by making sure Movable != true

This way it could also come in handy for staff created runebooks that are locked down in the world.

2. Allows the Summoners Kilt to be repaired.
A lot of the Gargoyle clothing items need to be fixed.

### DIFF
--- a/Scripts/Gumps/RunebookGump.cs
+++ b/Scripts/Gumps/RunebookGump.cs
@@ -332,7 +332,7 @@ namespace Server.Gumps
             }
             else if (buttonID == 1) // Rename book
             {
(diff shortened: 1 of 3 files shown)


kevin-10:
Put a comment on your commits so its easier to see them.

Tasanar:
What do you mean?

PR #4313, opened 2019-03-15
Bug Fixes

Tasanar:
1. Fixes a cliloc error.
2. Fixes an issue where all engraving tools can be used once the gump is up, even if the item is not in the players backpack.


Tasanar:
I would like to add that my changes were tested and confirmed to work.

PR #4348, opened 2019-04-01
Kotl Fixes

Tasanar:
After pouring over a few YouTube videos, players got a lot more drops. After running a few simulations the AVG was 2 - 3 drops. Should be double. It takes a lot of work and or money to get to the simulation.

2nd - Added and fixed various missing checks to the Automaton.

### DIFF
--- a/Scripts/Items/Functional/Automaton/KotlAutomaton.cs
+++ b/Scripts/Items/Functional/Automaton/KotlAutomaton.cs
@@ -173,13 +173,13 @@ namespace Server.Items
         public override double WeaponAbilityChance { get { return 0.33; } }
 
         public override bool IsScaredOfScaryThings { get {  return false;  } }
-        public override bool IsScaryToPets { get { return true || !Controlled; } }
+        public override bool IsScaryToPets { get { return !Controlled; } }
         public override FoodType FavoriteFood { get { return FoodType.None; } }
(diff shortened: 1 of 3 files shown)


kevin-10:
You need to put a subject line on your commit so i can click it to see the changes.

Scripts/Items/Functional/Automaton/KotlAutomaton.cs
```diff
@@ -173,13 +173,13 @@ public KotlAutomaton(Serial serial)
         public override double WeaponAbilityChance { get { return 0.33; } }
 
         public override bool IsScaredOfScaryThings { get {  return false;  } }
-        public override bool IsScaryToPets { get { return true; } }
+        public override bool IsScaryToPets { get { return true || !Controlled; } }
```
kevin-10: remove the 'true' so its says:
public override bool IsScaryToPets { get { return !Controlled; } }

Tasanar:
Alright so I made your changes and did some more digging.

Seems part of the code was just copied from a Golem. While a Golem can be barded, once controlled, one of the KOTL Automatons special abilites is that it is immune to bards.

Since it also did not exist pre AOS I dropped that check all together.

Golems were missing the "Scary to pets" after Controlled.

PR #4353, opened 2019-04-03
Additions:

kevin-10:
Astronomy https://uo.com/wiki/ultima-online-wiki/gameplay/astronomy/

Don't forget to configure the dates for the AnniversaryGiver21st in AnniversaryGiver21st.cs. The date set needs to be AFTER the current date to ensure each player gets a gift.

Bug Fixes
- Fixed an issue where certain addons were not being created properly in world decorate. Next server start will replace addons that were deleted
- Aegis of Grace (circlet version) is now meddable
- Flesh Ripper now has Mage Slayer, per EA
- Sand is now stackable and supports int32 constructor Arg
- City Trade Deals that increase stats not work properly
- Createing characters in Enhanced Client now gives the proper starting stats and skills
- Removed invalid FacialHairItemID from Garamon
- Fixed NAN error when calculating MinTameChance when skill was set to zero
- Tiger Cub now takes damage from their Master
- Tiger Cub Statuette can now only be used once. Until this is verified on EA, this will eb how it remains
- Tiger Cubs can now be trained up to 3 slots
- Added EOWMU (Spelling?) to pet training table. It will be like all other vanity pets.
- Remove Trap now has a skill variance of 10 (used to be 30) for skill gains, per EA


kevin-10:
Found an issue need to fix.

PR #4365, opened 2019-04-10
fixs.

Argalep:
* Stat Reward fix.
* SilverSapling fix.
* GemOfSalvation fix.

### DIFF
--- a/Scripts/Mobiles/PlayerMobile.cs
+++ b/Scripts/Mobiles/PlayerMobile.cs
@@ -3758,14 +3758,6 @@ namespace Server.Mobiles
             if (Criminal)
                 BuffInfo.RemoveBuff(this, BuffIcon.CriminalStatus);
 
-            if (Backpack != null)
-            {
-                GemOfSalvation gem = Backpack.FindItemByType<GemOfSalvation>();
-
-                if (gem != null)
-                    Timer.DelayCall(TimeSpan.FromSeconds(2.0), new TimerStateCallback<PlayerMobile>(gem.Use), this);
-            }
-
             DropHolding();
 
             if (Core.AOS && Backpack != null && !Backpack.Deleted)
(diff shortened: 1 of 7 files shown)


Scripts/Mobiles/PlayerMobile.cs
```diff
@@ -3755,6 +3758,14 @@ public override bool OnBeforeDeath()
...
                 BuffInfo.RemoveBuff(this, BuffIcon.CriminalStatus);
 
+            if (Backpack != null)
+            {
+                GemOfSalvation gem = Backpack.FindItemByType<GemOfSalvation>();
+
+                if (gem != null)
+                    Timer.DelayCall(TimeSpan.FromSeconds(2.0), new TimerStateCallback<PlayerMobile>(gem.Use), this);
+            }
+
```
kevin-10: Can we invoke an event for this (PlayerDeathEvent)? Also, should we be using Persistence to serialize all these things that very few players will use, but will cause memory usage on all instances?

Scripts/Services/VeteranRewards/StatRewardGump.cs
```diff
@@ -7,40 +7,31 @@ namespace Server.Engines.VeteranRewards
...
         }
 
         public override void AddGumpLayout()
         {
-            AddBackground(0, 0, 291, 173, 9200);
-            AddImageTiled(5, 5, 281, 140, 2624);
-
-            //AddHtmlLocalized(7, 7, 277, 126, 1076664, 0x7FFF, false, false);
-            /*<B>Ultima Online Rewards Program</B><BR>Thank you for being part of the Ultima Online community for over 6 months. 
-             * As a token of our appreciation, your stat cap will be increased. */
```
kevin-10: ServUO and RunUO use 1 month = 1 year for vet rewards. This is why the change was made to teh stat rewards in the first place.

PR #4379, opened 2019-04-26
Bug fixes

kevin-10:
- where skills at absolute minimum would have 0 chance to succeed.


kevin-10:
I don't know what you mean. Nothing has changed, except at minimum skill you now have minimum chance to succeed. Before at minimum skill, you failed every time since Utility.RandomDouble will never equal 0.

PR #4395, opened 2019-05-13
Update PetTrainingHelper.cs

uomamoru:
For enable add skill of "Animal Taming" by accessories when training pet.


Tasanar:
I'm going to let Dexter look at this first.

uomamoru:
> I'm going to let Dexter look at this first.

Thank you. ;)
Could you check with another pull-request #4396 together?
Regards.

kevin-10:
Thank you!  I removed the comment, we don't need that.

PR #4396, opened 2019-05-13
Update BaseCreature.cs

uomamoru:
Fix for need too big skill (108.0 from level 1) to allow training pet.

### DIFF
--- a/Scripts/Mobiles/Normal/BaseCreature.cs
+++ b/Scripts/Mobiles/Normal/BaseCreature.cs
@@ -760,7 +760,7 @@ namespace Server.Mobiles
                     double level = currentControlSlots - ControlSlotsMin;
                     double levelFactor = (double)(1 + (ControlSlotsMax - ControlSlotsMin)) / minSkill;
 
-                    current = Math.Ceiling(Math.Min(108.0, minSkill + (minSkill * ((levelFactor * 7) * level))));
+                    current = Math.Ceiling(Math.Min(108.0, minSkill + (minSkill * ((levelFactor * .7) * level))));
                 }
             }
 


kevin-10:
I will have to check this when i get home, but first glance tells me this will royally F things up.

uomamoru:
> I will have to check this when i get home, but first glance tells me this will royally F things up. If you could be more specific with what this is supposed to fix, that would be a help.

Thank you for your reply.
I'll show an example what's happen by this code.

e.g.) A unicorn has below properties:
ControlSlotsMin = 2
ControlSlotsMax = 5
currentControlSlots = 3 (means 1st training at this time)
minSkill = 95.1

double level = currentControlSlots - ControlSlotsMin;
    // level = 1

double levelFactor = (double)(1 + (ControlSlotsMax - ControlSlotsMin)) / minSkill;
    // levelFactor = 0.04206

current = Math.Ceiling(Math.Min(108.0, minSkill + (minSkill * ((levelFactor * 7) * level))));
    // minSkill + (minSkill * ((levelFactor * 7) * level)) = 123.1 (is too big.)
    // current = 108.0

The correct range of required skill is 95.1 to 108.0 with gradually rise at each step for training any pet.
But this code means 108.0  at all time for training unicorn.

If "* 7" should be "* 0.7", "current" will become 97.0. It's better.

And I know another bug that required skill on Gump of AnimalLore is not match when 1st training.

kevin-10:
This will need to be looked into. The formula I made was from actual testing on EA, but of course, I did not test all variations of creatures.

uomamoru:
OK, I see.
Anyway I'll tell you another example, to start training a horse, required skill is 51.0 at currently code. It's too high.
Thank you again.

kevin-10:
Then whats it supposed to be?

uomamoru:
> Then whats it supposed to be?

A horse? I checked that I could start training a horse with AnimalTaming is 29.2 on EA just now.

kevin-10:
You're not looking at it right. Once you train it, it bumps it's taming requirement up. So, in order to do the first round of training, you have to have the next levels minimum requirements in order to train. I have found a stratics post that normalizes the taming requirement, so the current formula is wrong, as well as yours.

So, the way its supposed to work:

Anytime you apply any training, your pets control requirement increases by 2.4. Anytime you move to the next level of training, the requirement increases by 21.2. Once I verify this on EA, I will implement it. Closing this PR.

uomamoru:
Thanks!
I meant "Can start training" is "Appeared the button for training on Gump of AnimalLore" on EA.
I'm looking forward to the update. ;) Thank you!
Regards.

PR #4400, opened 2019-05-16
Update PetTrainingHelper.cs

uomamoru:
We can not see the button on gump of AnimalLore for many pets to start training.
e.g.) Why do we need high skill of Animal Taming 117.1 (95.1+22) to start training (to see the button) a unicorn even now?


kevin-10:
Fix is in. If you don't understand the changes your making, please report it as a bug and I'll fix it. THis system is complex as you can see. But yea, I overlooked a small detail and will have a fix.

uomamoru:
Ok, I appreciate your kindness.

PR #4401, opened 2019-05-16
Update BaseCreature.cs

uomamoru:
I can not understand why this formula.
And even if there should be "MinTameSkill +", it's too high for great pets. e.g.) A unicorn grew up to slot 5 (from 2) become 159.1!


kevin-10:
This should not have been called unless you had an older version. None the less, I overlooked and didn't add a cap of 108.

uomamoru:
Oh! no, I meant your formula gives SAME values to ANY pet (has same slot) even if it's latest version.

PR #4436, opened 2019-06-13
Pet stat gain (EA Tested)

### LINKED ISSUE #4435: Pet stat gain
Pets should still be able to stat train up to 125 via combat even after Pet Training was introduced on official. I've tested this on multiple pets on sonoma.
Suggested fix in PR: https://github.com/ServUO/ServUO/pull/4436

ShadwDrgn:
Pets should still stat gain from combat even with PetTraining
Fixes: https://github.com/ServUO/ServUO/issues/4435

### DIFF
--- a/Scripts/Misc/SkillCheck.cs
+++ b/Scripts/Misc/SkillCheck.cs
@@ -481,14 +481,7 @@ namespace Server.Misc
 
             if (from is BaseCreature && ((BaseCreature)from).Controlled)
             {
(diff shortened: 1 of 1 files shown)


Tasanar:
So why waste training points on a stat?

ShadwDrgn:
Combat training goes to 125. Pet training goes to 700 str 150 Dex. Not sure
on int but much more than 125.

PR #4444, opened 2019-06-26
Make combat mastery Upkeep EA accurate

ShadwDrgn:
Fixes combat training inaccuracy


Scripts/Spells/Skill Masteries/CombatTraining.cs
```diff
@@ -433,34 +427,17 @@ public ChooseTrainingGump(Mobile caster, BaseCreature target, CombatTrainingSpel
 
             AddHtmlLocalized(20, 20, 150, 16, 1156113, Hue, false, false); // Select Training
 
-            int y = 40;
```
kevin-10: Why are these being reverted?

PR #4479, opened 2019-07-29
Fixes Multiple Issues with Whips:

Tasanar:
1. Virtue Shield can now be repaired.
2. Whips can now be imbued/reforged, use the correct specials, have the correct itemids and names, and can be correctly used by gargoyles.

https://trueuo.com/index.php?threads/whips-here-are-not-accurate-weapon-specials-are-wrong-fencing-is-misnamed.1729/


kevin-10:
Are these craftable?

Tasanar:
No not yet. Just event item.

PR #4485, opened 2019-08-08
Slight clean up to Mobile.cs and PlayerMobile.cs

Tasanar:
Code was duplicated in two spots. The Lord/Lady code and guild code. The Guild code in the core was outdated. Could probably use a total re-write since PlayerMobile gets their titles from the Title Menu now.


Tasanar:
Newest update removes Commands no longer used as well as cleans up some old old stuff that was no longer used. Lets bring ServUO into 2019.

PR #4487, opened 2019-08-08
Various bug fixes / code cleanup

Tasanar:
1. Darts now have the correct skill values when crafting. Tested on EA.
2. Removes various custom code hiding in the project.
3. Removes depreciated items and commands.
4. Merges duplicate code found in both PlayerMobile and Mobile.cs (Still needs to be looked at more)


Scripts/Mobiles/PlayerMobile.cs
```diff
@@ -5827,6 +5820,7 @@ public override void AddNameProperties(ObjectPropertyList list)
             }
 
             suffix = ApplyNameSuffix(suffix);
+			string name = Name;
```
kamronbatman: indentation

PR #4489, opened 2019-08-10
Fixed BankerGump issue

Tasanar:
Fixed an issue where players could open an unlimited number of BankerGump and the Help gump inside of BankerGump.cs


Scripts/Items/Artifacts/Tools/JacobsPickaxe.cs
```diff
@@ -70,18 +70,14 @@ public JacobsPickaxe(Serial serial)
...
             writer.WriteEncodedInt(0); // version
         }
 
         public override void Deserialize(GenericReader reader)
         {
             base.Deserialize(reader);
-
             int version = reader.ReadEncodedInt();
-
-            _Instances.Add(this);
```
kevin-10: Why is this being removed?  They won't get checked during the timer.

Tasanar: Ah I thought it was just an old Deser to fix older bugged ones.

PR #4502, opened 2019-08-28
Fixes a cliloc error

Tasanar:
Fixes a cliloc error reported on the forums.

### DIFF
--- a/Scripts/Services/Craft/DefCartography.cs
+++ b/Scripts/Services/Craft/DefCartography.cs
@@ -103,7 +103,7 @@ namespace Server.Engines.Craft
             AddRes(index, typeof(TreasureMap), 1073502, 1, 1073503);
             AddResCallback(index, ConsumeTatteredWallMapRes);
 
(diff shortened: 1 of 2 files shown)


Tasanar:
Removed un-needed code in PaladinSpell

The saying of the mantra is already defined in Spell.cs

PaladinSpell was the only magic school to have an override, that is not needed.

kevin-10:
This spell mantra may have issues that come up later, or at least inconsistencies with EA unless verified. It looks like the override was to make the spell mantra the same color regardless of speech hue, and whatever MessageType does which can have effects on journal logging etc.

The difference, Spell.cs:

m_Caster.PublicOverheadMessage(MessageType.Spell, m_Caster.SpeechHue, true, m_Info.Mantra, false);

In PaladinSpell:

Caster.PublicOverheadMessage(MessageType.Regular, 0x3B2, MantraNumber, "", false);

Tasanar:
So on real UO too, Chiv is the only one that acts differently?

kevin-10:
Im not sure. We'd have to test.

PR #4515, opened 2019-10-15
Add SetMutateAction helper function to the craft system.

DarkLotus:
Enables people to set a callback action to mutate an individual craft entry without needing to modify the base Type.

Example usage
``` 
index = AddCraft(typeof(EnchantedApple), 1073108, 1072952, 60.0, 85.0, typeof(Apple), "Red Enchanted Apple", 1, 1044253);
SetMutateAction(index,(player, item, tool) => { item.Hue = 55; });

index = AddCraft(typeof(EnchantedApple), 1073108, 1072952, 60.0, 85.0, typeof(Apple), "Green Enchanted Apple", 1, 1044253);
SetMutateAction(index,(player, item, tool) => { item.Hue = 99; });
```

Etc, access to the crafted item, the crafter and the tool used is provided in the callback.

### DIFF
--- a/Scripts/Services/Craft/Core/CraftItem.cs
+++ b/Scripts/Services/Craft/Core/CraftItem.cs
@@ -89,6 +89,8 @@ namespace Server.Engines.Craft
         public bool NeedWater { get; set; }
         public int ItemHue { get; set; }
 
+        public Action<Mobile, Item, ITool> MutateAction { get; set; }
+
         public void AddRecipe(int id, CraftSystem system)
         {
             if (Recipe != null)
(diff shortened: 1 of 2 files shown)


kevin-10:
I think we already have something like this.

kevin-10:
See:
        /// <summary>
        /// this delegate will handle all crafting functions, 
        /// such as resource check, actual crafting, etc. 
        /// For use for abnormal crafting, ie combine cloth, etc.
        /// </summary>
        public Action<Mobile, CraftItem, ITool> TryCraft { get; set; }

        /// <summary>
        /// this func will create complex items that may require args, or other
        /// things to create that Activator may not be able to accomidate.
        /// </summary>
        public Func<Mobile, CraftItem, ITool, Item> CreateItem { get; set; }

        public Func<Mobile, ConsumeType, int> ConsumeResCallback { get; set; }

DarkLotus:
The above "CreateItem" is for when you want to entirely override the created item though, this helper just mutates the created item.

kevin-10:
Ahh, yes you are correct. I will re-open and approve.

PR #4541, opened 2020-02-02
Fixes JadeArmband and puts it in the correct location.

TrueUO:
Fixes JadeArmband and puts it in the correct location.


TrueUO:
I have no idea why this keeps failing.

PR #4547, opened 2020-02-11
Designer Packet Handling

Vorspire:
Add placeholder packet handler for new designer packet 0x0A


kevin-10:
So...how do we find out what these do? I suspect 0x0F, which is a response for the detailed house foundation SendInfo packet, is used for clearing out non-visible items in houses.

PR #4566, opened 2020-02-24
Can only open moongate gump if within 3 steps.

TrueUO:
Confirmed on UO
- Must be within 3 steps to attempt to open the public moongate gump.
- Currently, on ServUO you could double click it from anywhere and get the gump.


kevin-10:
If CanUseGate returns false, the gump never sends:

https://github.com/ServUO/ServUO/blob/0e0fd72974d033e480fad81468db312f8a5f7179/Scripts/Items/Functional/PublicMoongate.cs#L218

PR #4569, opened 2020-02-27
Moongate Fix

kevin-10:
- You have to be 1 tile to double click it. All other range checks are correct.


kevin-10:
Can someone explain to me why all these commits, all of which have been merged to the master repo, keep appearing on my PR's?

@Vorspire

PR #4595, opened 2020-03-15
* add ImprovedRockHammer

Argalep:


Scripts/Services/Harvest/Mining.cs
```diff
@@ -221,6 +205,14 @@ public override Type GetResourceType(Mobile from, Item tool, HarvestDefinition d
 
                 PlayerMobile pm = from as PlayerMobile;
 
+                if (tool is ImprovedRockHammer)
```
kevin-10: Can we just do this:
double chance = tool is RockHammer ? 0.50 : tool is ImprovedRockHammer ? 1.0 : 0.15;

Or have ImprovedRockHammer derive from rock hammer and make an overridable property to define bonus chance. That way if there are more introduced, its easier to handled.

PR #4602, opened 2020-03-21
Circuit Trap Traning Kit fix

ravsz:
The Circuit Trap Traning Kit with 16 fields has two unsolvable paths


TrueUO:
Looks good thanks! Going to have Dexter take a look too.

kevin-10:
Thank you!

PR #4614, opened 2020-03-24
Fix commodity deed

ravsz:
Commodity deed description do not contains boards color


TrueUO:
Thanks for the help @ravsz

PR #4620, opened 2020-03-25
Final fix.

TrueUO:
Final Fix

### DIFF
--- a/Scripts/Mobiles/AI/BaseAI.cs
+++ b/Scripts/Mobiles/AI/BaseAI.cs
@@ -2879,62 +2879,52 @@ namespace Server.Mobiles
 					{
 						continue;
 					}
-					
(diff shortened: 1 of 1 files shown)


kevin-10:
So its not doing an IsFriend check?

TrueUO:
I really do not think it is needed when you look at the whole method from before;

The IsFriend was in the fraction brackets, so when we deleted the whole method it was only checking again, sanity?, that still !is not this or that, when now it does not need to. No idea if that makes sense but if you look at BaseAI from two days compared to this it seemed the right answer.

TrueUO:
Argalep can you look at this too?

kevin-10:
Gotcha. Im at work and don't like looking at old versions through github.

PR #4628, opened 2020-03-26
Cleanup

kevin-10:


Scripts/Mobiles/Normal/BaseCreature.cs
```diff
@@ -3813,7 +3686,7 @@ public int ControlSlots
         #endregion
 
         public virtual bool AutoDispel { get { return false; } }
-        public virtual double AutoDispelChance { get { return ((Core.SE) ? .10 : 1.0); } }
```
TrueUO: Shouldn't this be .10 and not 1.0?

PR #4647, opened 2020-03-31
Update BaseGalleon.cs

ravsz:
It's very likely that this will fix the problem with galleon crash bug


TrueUO:
Yeah, not going to merge something that is not needed. I am not getting this crash at all. If it is only one person could it possibly be a bad merge?

ravsz:
If you could review changes that was added there don't change any logic.
It's only three lines and they only improve the iteration method thru the list from which elements are removed

Sure, it's understandable if you dont want merge for now, but anyway if issue occurs again then you could pull out potential fix of the freezer, enjoy :)

PR #4661, opened 2020-04-02
Revert "- remove all old OnSingleClick functions"

kevin-10:
Reverts ServUO/ServUO#4660


kevin-10:
Reverting, this broke context menus from mobiles and items.

PR #4690, opened 2020-04-09
SecretChest fix.

Argalep:


Scripts/Items/Containers/LockableContainer.cs
```diff
@@ -300,6 +300,17 @@ public override bool TryDropItem(Mobile from, Item dropped, bool sendFullMessage
 
         public override bool OnDragDropInto(Mobile from, Item item, Point3D p)
         {
+            if (this is SecretChest sc)
```
kevin-10, line 303: Can we override OnDragDropInto in SecretChest.cs?

Argalep: I derived the SecretChest class from LockableContainer. There is a problem because I use the Locked parameter in common.

kevin-10:
So this wont work in SecretChest.cs?

        public override bool OnDragDropInto(Mobile from, Item item, Point3D p)
        {
            if (Locked && from.IsPlayer() && !CheckPermission(from))
            {
                from.SendLocalizedMessage(1151591); // You cannot place items into a locked chest!
                return false;
            }

            return base.OnDragDropInto(from, item, p);
        }

Argalep:
LockableContainer.cs also doesn't work because the code below is.

`public override bool OnDragDropInto(Mobile from, Item item, Point3D p)
        {
            if (from.AccessLevel < AccessLevel.GameMaster && m_Locked)
            {
                from.SendLocalizedMessage(501747); // It appears to be locked.
                return false;
            }

            return base.OnDragDropInto(from, item, p);
        }`

Issue #4686, opened 2020-04-09
Some NPCs do not handle Honesty items correctly


djredman99:
Creating this issue for feedback before submitting a PR for a fix.

In several cases, certain NPCs that "own" an Honesty item do not accept the item as an Honesty item.  Armourer's for instance, accept any weapon/armor item and if handed an Honesty item they are identified as an owner of, they do not grant an increase in Honesty, nor respond that the item is theirs.  Instead, they take their default behavior (in BaseVendor.cs) and say 'thanks' and some fame is awarded to the player mobile.

Another example would be the Silver Trader, they simply will not accept the Honesty item.

The reason for this is that the Honesty logic is happening in the BaseCreature class, and these certain NPCs are handling the event at either the BaseVendor or a more specific class level.  They never pass the logic to BaseCreature.cs. 

This would be easy to fix by doing a check to see if the item dropped on the NPC has a valid HonestyItemSocket at the higher class level and passing it to the BaseCreature class for the Honesty logic if it does.

kevin-10:
Sure, push the PR and we'll go from there. Nice catch!

djredman99:
I'll be testing a fix this evening and should have a PR soon.  Thanks!

djredman99:
Fixed and merged in with 02f76827d00dcb2e90487f8d1dd643a662ce81de

PR #4697, opened 2020-04-11
Update to Honesty item turn-in for NPCs

djredman99:
Regarding Issue #4686

I moved the logic for the Honesty check into BaseVendor since only vendors can be assigned as an owner of an item, so it had no reason to be in BaseCreature.

I tested with Armorer, blacksmith's, and Silver Trader.  They now accept the Honesty item correctly.  Tested handing items to many mobile types , both honesty items, gold, and random objects, just to be sure BaseCreature logic is still good.


kevin-10:
Well done, nice catch and thank you!

TrueUO:
Yeah much better being housed in BaseVendor over BaseCreature. Thanks for your contribution!

PR #4707, opened 2020-04-12
Escortable Cleanup

TrueUO:
This removed the outdated pre-ML Escortables.

All these Escortables had two versions and the 2nd version still exists in Quests/Escortables.

All spawn files have been upgraded for new shards.

Existing shards will have to update their existing spawners. Just clean up the ones that appear in the badspawn log.


kevin-10:
I'm pretty sure these still exist on EA. Do the new escortables keep the titles "Seeker of adventure" and "Noble?"

TrueUO:
Yup 100% We had two copies of both. Check out Scripts/Quests = Escortables

If you even look at EscortableMage somehow two items with that same name existed.

TrueUO:
I even converted all the spawns over to use the new values. Though existing shard owners will have to update on their own. (Already did it on Heritage, took 2 minutes)

Issue #4699, opened 2020-04-11
Honesty items sometimes incorrectly assign owners


djredman99:
When looking into issue #4686, I noticed that it takes 1/10 of the vendors and tries to find a matching owner in the same region as where the lost item is assign as "from".  However, after a number of attempts equal to 1/10 of the vendor count, it gives up trying to find a match and just assigns the last vendor it looked at.  This means you could have an owner from Moonglow Trammel while the lost item is from Yew Felucca, sending anyone using Forensic ID on a lost cause hunting for the owner in Yew Fel.

I have hunted Honesty items in game and at times, I search the entire city, and cannot find the named owner.  It now makes sense to me as to why.

Just making sure this is not by design (intentionally making honesty farming harder), let me know, and I will PR a fix.

djredman99:
I have confirmed this is happening.  I generated a lot of honesty items and after only the ~20th item, I had an item that was from Moonglow, but assigned an owner that was in Yew.

PR #4709, opened 2020-04-12
Lost Honesty item owners are now set correctly according to lost item…

djredman99:
This is in regards to issue #4699

Lost Honesty item owners are now set correctly according to lost item region.  

I modified the code to generate a list of all matching vendors in the region that was assigned to the honesty item and randomly select a vendor from that list.

Tested against latest master code and it was always assigning vendors correctly now instead of a "best effort" attempt that often failed to match the two correctly. 

I did provide a fail-over in case a server owner does not currently have any vendors generated in a specific area.  Instead of complicating things with a null owner, as I wasn't sure what implications this would have, I randomly assigned a vendor.


kevin-10:
Thanks!

PR #4716, opened 2020-04-13
Update CircutTrap

TrueUO:
The size of a collection and the length of an array are always greater than or equal to zero. So testing that a size or length is greater than or equal to zero doesn't make sense, since the result is always true. Similarly testing that it is less than zero will always return false.


kevin-10:
Should you remove the = or remove the entire statement?  Does the trap behave properly still?

PR #4735, opened 2020-04-14
Fix for RunicAtlas

djredman99:
Runic Atlas was getting stuck as immovable because it was failing to reopen the gump after a server response for Renaming the book (or canceling the response).  This left the user as an 'opener' in the books list of openers therefore acting as if it were in use.  The only solution in game was to log out and back in to make it movable again.

Additionally, you could open as many Runic Atlases as you wanted at the same time.  Code now makes it operate exactly like a Runebook where opening one will close any other that are currently open.


TrueUO:
Thanks! Are you on Discord?

djredman99:
Yes I am! I will contact you.

TrueUO:
Thanks again. https://github.com/ServUO/ServUO/pull/4741

PR #4736, opened 2020-04-15
Correcting Maritime Cargo values

djredman99:
Correcting Maritime Cargo values for the CityLoyaltySystem as well as correcting the high end cap for both doubloon turn-in with The Black Market Merchant and the CityLoyaltySystem.

When turning the cargo into a non-matching city, the value should be halved.  This was not happening in the code.

Referencing data from [UO.com](https://uo.com/wiki/ultima-online-wiki/combat/pvm-player-versus-monster/rising-tide/) and [UO-cah](https://www.uo-cah.com/pub-104/rising-tide-guide)


TrueUO:
We have having some issues with TravisCI give me a second going to open and push a PR from my end with all your fixes.

TrueUO:
Thanks again https://github.com/ServUO/ServUO/pull/4741

Issue #4708, opened 2020-04-12
Books are not saving correctly in the new client


djredman99:
The new client has changed how books are handled. Tested with client 7.0.85.15

- Book pages are now comprised of 10 lines (previously 8)
- Packets sent from the client are now 1 packet per page (previously one packet per book)

djredman99:
How would you want to handle this @kevin-10, @TrueUO .  I would assume the best solution would be to check for which client version is connected and handle the packet accordingly?  The logic to handle the new packet should be easy enough to do.

TrueUO:
Since we only support the newer clients, and it would be broken using older ones regardless, probably okay to just patch it for current, no?

djredman99:
Sounds good to me. I certainly understand the desired to simplify and would be ok with it.  I’m sure that is how Broadsword operates official servers.  If you don’t patch, expect issues.  Wasn’t sure how much backwards compatibility you are trying to preserve.

djredman99:
I should have this done next day or two.

TrueUO:
Awesome. Yeah since this fixed a UO issue stemming from the release of the EC and how books entered there had issues displaying in the CC, and vice versa, any client that is relatively new will benefit from this. While older clients can use the current method. If you have EC questions @Argalep is the go-to guy.

djredman99:
@TrueUO  “While older clients can use the current method”

So wait, support the older client or no?

TrueUO:
no

PR #4755, opened 2020-04-18
Update to books for new client packet

djredman99:
Relate to issue #4708

Tested with Classic Client and Enhanced Client
Tested regular books and XmlSpawner Book entries from XmlSpawnerGump


djredman99:
I think I am going to close this PR and try to push it again.  I setup Travis CI to run on my fork, but not sure it will report it for this PR since it was generated before the policy was set on the fork.

The other solution is to remove the "Travis CI - Branch" policy to let this go through, turn back on and see if we get it reported on the next one.

PR #4787, opened 2020-05-26
ServUO_Pub57 - Fixed Server Crash when teleporting people out in PeerlessAltar.cs

miguelmoraleda:
The issue has been described [in this post](https://www.servuo.com/threads/server-crash-when-server-is-moving-players-out-of-bedlam-when-boss-is-death.12620/#post-75581)

Tasanar has provided with this solution and they are merging to master, I'm just sending a PR to keep ServUO_Pub57 in good conditions.


TrueUO:
Thanks a lot, you rock!

This is the kinda stuff we need. We will open other branches of ServUO if people want to keep them up to date, we highly encourage this type of support!

PR #4790, opened 2020-06-04
FIXED server crash when using old tiledata

emilieandthebee:
FIXED: the server was crashing when using old versions of tiledata for HouseTeleporterTile as it is hardcoded.
You may check all all the posts about this error by searching "HouseTeleporterTile" in servuo forum.
The fix is obvious.


TrueUO:
Thanks a lot! Looks good.

PR #4801, opened 2020-06-26
Various Fixes

Vorspire:


kevin-10:
Thanks for looking into the stat timer. I've been hearing a few complaints lately and didn't get a chance to look into it. One question. @Argalep mentioned timers do get caught up in garbage collection, even if there are no more pointers for them. Is this something we should look into as well as a potential memory leak?

PR #4832, opened 2020-08-22
Do not dynamically compile Scripts

jaedan:
Scripts are now compiled into Scripts.dll and linked with Server.dll to form ServUO.exe. There is no compilation at run time.

The build scripts will still function as normal but now require the 'dotnet' tool which is available in the .NET core SDK. The project itself still builds against .NET 4.8. The build scripts are now just simple wrappers around the following command:

`dotnet build -c [Debug|Release]`

When deploying ServUO to a server you no longer need to copy all of the source code. You only need to move the DLLs and the EXE. This is because everything is precompiled.

This change also prepares the code to eventually move to .NET core 3.1 or .NET 5 by updating the csproj files to the new format. There are a few legacy areas of code that need to be updated to become compatible, but it's a relatively small task.


Vorspire:
> @Vorspire any idea why Travis is getting caught up?

I'm not sure. Perhaps Travis doesn't like certain actions like the force-push?
@jaedan would know better than I 💯

jkachhad:
Why not just merge Scripts/Server into a single project and compile into a single binary?

PR #4833, opened 2020-08-23
Update BaseVendor.cs

SaschaKP:
This will remove the entity after it has been completely bought from the vendor, this will fix the error from a client point of view, where the item will remain on the vendor "container" until the next containercontent update.

The reason for this is simple, after the item is being bought, the displayed entity will be different from the bought one, so all the 'remove' and 'Delete()' that affects the REAL item, won't affect the 'virtual' one, leaving spurious traces behind.


kevin-10:
I no longer have the ability to merge commits so somebody else is going to have to do it.

TrueUO:
Same here. I can not see anything on the Travis side here either.

PR #4850, opened 2020-09-18
Updates

Vorspire:


TrueUO:
Would that make the gift giving packages hand out yearly?

Vorspire:
Yes, which seems to be what every single shard edits the files to do anyway.

TrueUO:
Yeah, I would advise them against that.

If anything we should combine them all into one folder with one GiftGiver.cs as an example and shard owners can then deploy all those different gift packages at will. I will do that over at TrueUO and push it here for review.

TrueUO (changes requested):
GiftGiving. Leave dates as is for now to let shard owners decide how to deploy. 

Will work on condensing them all into one folder with one GiftGiver.cs

TrueUO (approved):

TrueUO (approved):

Issue #4778, opened 2020-05-07
Error running on macOS Catalina 10.15.4


lucasrencoret:
runned `brew install mono`, then `make` and I receive this error:
`error CS1644: Feature 'default literal' cannot be used because it is not part of the C# 7.0 language specification`

Is this code working for Mac?

kamronbatman:
I personally never got it working on OSX.

Fire-Dragon-DoL:
The same problem is happening on Ubuntu, this could be problematic because it makes sense to host an Ultima online server on an Ubuntu machine.

The problem is that mono has partial support for C# 7, but it doesn't mention 7.1.
The `default` keyword is part of 7.1.

kamronbatman:
Mono doesn't work well on Linux so I don't recommend it at all. You should convert ServUO to .net core and then you will have C# 8 support. Or wait until November/December for .Net 5

Fire-Dragon-DoL:
I'm happy to switch to.NET core, but the instructions mentioned mono, that's why I used it. Once I figure out how to compile it, I'll make a PR for the readme.
I'm not sure if the Makefile needs to be updated.

kamronbatman:
It is not about just compiling it, there will be _some_ code changes. I know the ServUO team would be happy to see a PR for .net core.

TrueUO:
Oh, absolutely...

Currently, with just myself, Dexter, and Argalep, we really don't have the know-how on how to do this correctly and efficiently while maintaining support for the shards currently using ServUO code.

PR #4869, opened 2020-09-30
PR: Makefile and ServUO.sln Update

pyro17:
The makefile was updated to run on linux,
the readme also includes the zlib, without that the server crashes too

and finally the changes to the ServUO.sln, with that you can grab the files, run vs, and instantly hit play instead of changing the start project everytime. Also set it use the Debug Profile when it is requested to run as such, for the Scripts project

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,11 @@
 MCS=mcs
 EXENAME=ServUO
 CURPATH=`pwd`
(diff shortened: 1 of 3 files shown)


pyro17:
#4778 should be fixed with this as well

Vorspire (approved):

Issue #4398, opened 2019-05-14
(mono) restart command causes crash followed by endless restart loop


Zavreyon:
This has occurred on both 56.0 release and a git clone taken today. Debian 9.9 x64 with mono 5.20.1.19. 

[Crash 14-5-2019-14-47-28.log](https://github.com/ServUO/ServUO/files/3178496/Crash.14-5-2019-14-47-28.log)

Occurs with no modifications to core, scripts, or config except Datapath to point to Muls. Happens when using the restart command both from server console and from ingame. 

14:47:28 Error:
14:47:28 System.IO.IOException: Invalid handle to path "/srv/ServUO/latest/[Unknown]"
  at (wrapper managed-to-native) System.Runtime.Remoting.Messaging.AsyncResult.Invoke(System.Runtime.Remoting.Messaging.AsyncResult)
  at System.Runtime.Remoting.Messaging.AsyncResult.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem () [0x00000] in <6649516e5b3542319fb262b421af0adb>:0
  at System.Threading.ThreadPoolWorkQueue.Dispatch () [0x00074] in <6649516e5b3542319fb262b421af0adb>:0
  at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback () [0x00000] in <6649516e5b3542319fb262b421af0adb>:0
14:47:28 Crash: Generating report...done
14:47:28 Crash: Backing up...done
14:47:28 Crash: Restarting...done

I do not appear to be the only one having this issue as indicated by these forum posts:
https://www.servuo.com/threads/on-unix-console-a-server-restart-creates-an-endless-loop.10964/
https://www.servuo.com/threads/restart-servuo-crash-linux.10991/

Zavreyon:
When using mono 6.0 preview release, I get a bit more in the crashlog, don't know if any of this helps or not:

[Crash 24-6-2019-15-53-44.log](https://github.com/ServUO/ServUO/files/3321614/Crash.24-6-2019-15-53-44.log)

kevin-10:
I highly doubt you'll find any mono support here.

Zavreyon:
I understand, I have since found a workaround using screen and systemd that allows server restarts without the loop occurring.

kamronbatman:
FYI this has not been fixed still.

TrueUO:
[...]

PR #4870, opened 2020-09-30
Fixed Mono Server Reboot

pyro17:
- Finally server on Linux can restart the server without having to manually close the server first.  #4398
- Also fixed the line break issue in the linux area in the readme, caused by me

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,8 @@ Run `Compile.WIN - Release.bat` for production environments.
 
 #### Ubuntu / Debian
 
-`apt-get install zlib1g-dev`
-`apt-get install mono-complete`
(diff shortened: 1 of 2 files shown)


kevin-10:
@Vorspire can you take a look bud, this is a bit out of my realm. Thanks!

Vorspire (approved):
Everything looks fine. 
I'm going to defer to @pyro17 on this and assume it's all working as I don't have a means to test this currently. I'm sure if there are any issues after, they can be resolved. It won't affect the build process for typical windows servers, so there's not a high risk of issue.

kevin-10:
Thank you sir!

PR #4873, opened 2020-10-02
Bug Fixes

TrueUO:
Fixed crash while enabling Kotl System
Added support for invoking commands without a mobile
Bolt of CLoth is no longer flipable with housing
Fixed issue with castle ladders
Chest guardians will now get Soul Bound tag during active Event
Fixed issue where physcial resist wasnt returning to normal value when disabling magic reflect spell

### DIFF
--- a/Scripts/Commands/Logging.cs
+++ b/Scripts/Commands/Logging.cs
@@ -86,21 +86,24 @@ namespace Server.Commands
 
             try
             {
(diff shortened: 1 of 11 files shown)


TrueUO:
Thanks @pyro17 for the quest item consumption fix.

kevin-10 (approved):

PR #4896, opened 2020-10-31
Fixed the missed mono issue

pyro17:
also refactored it to also skip the string splitting


TrueUO (approved):
Love this man, thanks!

PR #4899, opened 2020-11-15
Container Item Limit

Vorspire:
Fixed an issue where containers could go over their maximum item limit.


TrueUO:
Awesome, playing with this shortly.

TrueUO (approved):

PR #4917, opened 2021-01-15
Publish 58

Voxpire:
See individual commit notes for additional details.


kamronbatman (commented):
Looks great!

kevin-10 (dismissed):

Server/Items/SecureTradeContainer.cs
```diff
@@ -106,7 +106,7 @@ public override bool IsChildVisibleTo(Mobile m, Item child)
 		{
 			if (child is VirtualCheck)
 			{
-				return AccountGold.Enabled && (m.NetState == null || !m.NetState.NewSecureTrading);
+				return !AccountGold.Enabled || !m?.NetState?.NewSecureTrading != true;
```
kamronbatman, line 109: I think this probably should be:
```cs
!AccountGold.Enabled || m?.NetState?.NewSecureTrading != true
```

Either Account Gold is disabled, or the client does not support new secure trading.

LMK if that makes sense?

kamronbatman:
@Vorspire I would probably recommend gating the logging against a DEBUG constant because a malicious user could cause quite a bit of spamming on the console/logs. Thoughts on this?

kevin-10 (dismissed):

kamronbatman:
@kevin-10 you shouldn't approve since he hasn't pushed a fix for the typo.

kamronbatman (changes requested):
Need to fix the typo :)

Server/Item.cs
```diff
@@ -1089,8 +1074,17 @@ private void VerifyCompactInfo()
...
+					   || info.m_Items != null 
+					   || info.m_Bounce != null 
+					   || info.m_HeldBy != null 
+					   || info.m_BlessedFor != null 
+					   || info.m_Spawner != null 
+					   || info.m_TempFlags != 0 
+					   || info.m_SavedFlags != 0 
+					   || info.m_Weight != -1 
+					   || info.m_VisibilityList != null
+					   ;
```
kamronbatman, line 1087: Weird spacing format?

Scripts/Commands/Profiling.cs
```diff
@@ -213,7 +216,8 @@ public static void TraceExpanded_OnCommand(CommandEventArgs e)
                         "TempFlag",
                         "SaveFlag",
                         "Weight",
-                        "Spawner"
+                        "Spawner",
+						"VisList"
```
kamronbatman, line 220: Bad spacing here too? I think you are using tabs 🙈

Server/Item.cs
Vorspire: On purpose.... making things easier for others to merge in future.

PR #4955, opened 2021-08-07
ServerListMap: ServerList selection by matching CIDR of remote address

ptgeorge:
### Problem
Could not connect to a shard I created locally in a docker container (on a different machine).  The initial login would work, but then selecting a shard would cause the client to hang.  Also would not work over a SSH tunnel.

### Solution
1. Check remote address against a list of CIDRs / network
2. If a remote address lies within a CIDR / network, then use that one (do not check any other CIDRs / networks in the list)

The list is in Config/ServerListMap.xml.  XML seems to be the way ServUO handles any structured or multi-element configuration, which is why that approach was selected.


Scripts/Scripts.csproj
```diff
@@ -27,4 +27,8 @@
   <ItemGroup>
     <Reference Include="System.Web" />
   </ItemGroup>
-</Project>
\ No newline at end of file
+
+  <ItemGroup>
+    <PackageReference Include="IPNetwork2" Version="2.5.329" />
```
kamronbatman, line 32: I don't see this used?

kamronbatman:
Thank you for the contribution!

This looks good however essentially overrides the existing way of configuring the remote IPs to include specific CIDRs.

So two things, first maybe we can combine these ways of configuring somehow so they are unified?

Second, looks like your indentation is inconsistent. Let's try to clean up indents in the file as much as possible.

Scripts/Scripts.csproj
ptgeorge: The nuget package is called IPNetwork2:
https://github.com/lduchosal/ipnetwork

kamronbatman: I meant I don't see the nuget package used in the PR, unless I missed something?

ptgeorge:
> Thank you for the contribution!
> 
> This looks good however essentially overrides the existing way of configuring the remote IPs to include specific CIDRs.
> 
> So two things, first maybe we can combine these ways of configuring somehow so they are unified?
> 
> Second, looks like your indentation is inconsistent. Let's try to clean up indents in the file as much as possible.

It actually it meant to compliment, not replace.  For example, if there's a connection from a public address, the auto-detection could still be helpful and require less configuration from common users.  Many won't need a complex config.  It was a goal to maintain the existing behavior.  I'm happy to adjust the way it operates if there's a specific suggestion.

Sorry about the indents -- I'll clean it up.

Scripts/Scripts.csproj
ptgeorge: The package is called IPNetwork2 but the class is called IPNetwork within the System.Net namespace:
https://github.com/lduchosal/ipnetwork/blob/master/src/System.Net.IPNetwork/IPNetwork.cs

ptgeorge:
Spacing/indents cleaned up

sahensley:
Nice job! 👍 

I've tried for months on-and-off trying to get ServUO to run in a container and not hang when connecting with ClassicUO.  After I built ServUO based on the PR and set the mapping in ServerListMap.xml, connecting works as expected.

sahensley:
@kamronbatman Could this PR be revisited/reviewed again please? Do further changes need to be made?

So far this has been the only way I've been able to get ClassicUO to not hang while connecting if the server is run in a container.

kamronbatman:
> @kamronbatman Could this PR be revisited/reviewed again please? Do further changes need to be made?
> 
> So far this has been the only way I've been able to get ClassicUO to not hang while connecting if the server is run in a container.

I'll look into it when I get some time. Thank you. Cc @Voxpire if you have any thoughts on this.

Voxpire:
If it is integrated properly into the normal login API flow rather than being an invasive modification that requires a 3rd party package, then I'm all for it.

I don't include third party packages in ServUO for accessibility reasons.

kamronbatman:
> If it is integrated properly into the normal login API flow rather than being an invasive modification that requires a 3rd party package, then I'm all for it.
> 
> I don't include third party packages in ServUO for accessibility reasons.

I was able to accomplish this in MUO without a module. I don't remember if the code is specific to dotnet.

kamronbatman:
I looked up ModernUO code, and this is the flow:
Parse the CIDR into an address and length
```cs
var cidrParts = cidrText.Split('/');
IPAddress.Parse(cidrParts[0], out var cidrAddress);
int.TryParse(cidrParts[1], out var cidrLength);
```
Record the address and length.
Then when you need to determine the mapping:
`Utlity.IPMatchCIDR(cidrAddress, addressToCheck, cidrLength);`
If it matches, then use one of the configured addresses.

sahensley:
Kamron, thank you for the code snippet, that's great! :smile: 

@ptgeorge - Would you be interested in working on these changes?  If not, I can work on something in the coming week(s), free time permitting.

kamronbatman:
@sahensley I updated the PR, please test it.

.editorconfig
```diff
@@ -7,6 +7,8 @@
 
 #use hard tabs for indentation
 indent_style = tab
+indent_size = tab
+tab_width = 4
```
kamronbatman, line 11: I had issues with Rider, and this fixed it.

kamronbatman:
I'll add in some tests, and defer to @Voxpire for updates to the config setup.

sahensley:
@kamronbatman - It works as expected, awesome work :tada: Also a special thanks to @ptgeorge for the original PR, which helped me realize the problem was some network funkiness and not how I had the server configured.

**Testing Results**
With the network mapping in ServerListMap.xml, the character list loads as expected and I can log in with the character.

Without having the mapping in ServerListMap.xml: The client lists my shard, I click on it, then it goes hangs on "Logging into the World" indefinitely, which is the expected result.

ptgeorge:
Awesome 👏. Cool there was a way without a 3rd party library…I’m obviously not as familiar with this code base and related as some.

ShadwDrgn:
May I suggest an alternative to this:
                if (AlwaysPublic && _PublicAddress != null)
                {
                    localAddress = _PublicAddress;
                }
                else if (IsPrivateNetwork(localAddress))

in EventSink_ServerList with 
        public static readonly bool AlwaysPublic = Config.Get("Server.AlwaysPublic", false);
at the top. This way you can just set it to ALWAYS use your public hostname and use your router to resolve yourdomain.com to your correct lan ip while still having yourdomain.com publically resolve to your real public ip.

ShadwDrgn:
Just a heads up for anyone who cares about this issue still... The above pull request fixes this with less change and I've tested it both in lan and wan now seems to work great. :D

ptgeorge:
> Just a heads up for anyone who cares about this issue still... The above pull request (#5047) fixes this with less change and I've tested it both in lan and wan now seems to work great. :D

Doesn’t cover the same cases but glad something workable is merged in.  Probably works for most docker related cases 🤷

PR #4974, opened 2021-12-09
VendorSellCheckBuyPack

TrueUO:
Vendors currently check to see if items should be deleted from their BuyPack every time someone uses the Buy option.

`if ((item.LastMoved + InventoryDecayTime) <= DateTime.UtcNow) { item.Delete(); }`

The problem is in some rare situations someone may only be selling items to a vendor, never using the buy option. Especially if this vendor is in the middle of nowhere it adds to item count unnecessarily. Some users have reported finding hundreds of thousands of items on an NPC, probably due to people macroing crafting skills.

If someone only ever sells and runs the item count to hundreds of thousands of items, there is a lag/delay once someone does finally come along and use the Buy option as now the server is deleting them and the client itself is flooded by the increase of item count for a brief moment.

This simply adds another check that also triggers every time someone goes to sell items to make sure we are keeping item count low. :)

### DIFF
--- a/Scripts/Mobiles/NPCs/BaseVendor.cs
+++ b/Scripts/Mobiles/NPCs/BaseVendor.cs
@@ -1066,6 +1066,25 @@ namespace Server.Mobiles
                 return;
             }
 
+            Container cont = BuyPack;
+
+            List<Item> packItems = cont.Items;
+
+            for (int i = packItems.Count - 1; i >= 0; --i)
+            {
+	            if (i >= packItems.Count)
+	            {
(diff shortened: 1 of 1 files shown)


sircapted:
so how do they connect to the config files for restock..
i have been hoping for a vendor that self sustains.. sells what he buys..  dreams.. that it could be a config option..hmmm
 and   the issue of pricing items  as to value by attributes with all items id'd by default.
say i can sell a repond or a  cheap item and vendor pays same price for both?

PR #4976, opened 2021-12-21
Update BaseQuest.cs

ruaduck:
Need to allow escortable to walk when quest is accepted (they are marked as can't walk when they are part of one of the Camps as a Prisoner)


Voxpire:
The update for escorts is ready in p58-wip, DDNS support can be submitted separately if done against the p58-wip branch, as that has extensive updates incoming for server configuration.

PR #4978, opened 2021-12-24
Removing hard coded caps to skill checks

noodleses:
Removing hard coded caps to skill checks limiting the skill gains to a certain amount


Voxpire:
Sometimes they need to be limited because the success chance should remain at 100% for skills above 100 in some contexts.
This is a little too volatile to merge, but I still appreciate the effort!

PR #4979, opened 2021-12-25
Adding new config file and values for carry weight

noodleses:
Adding new config file and values for carry weight values and calculations


Kylezzz101:
this is really nice

Voxpire:
Another great update!

PR #4982, opened 2021-12-31
Adding new auto looting of gold feature.

noodleses:
Adding a new feature under the category of "quality of life improvements".  I am aware that some client UIs have some version of this functionality built for them but it being on the server this is a common way to allow for loot sharing across all that were involved in the kill and even to all party members. This will therefore be done no matter the client used by all players.

I realized some people really want quality of life features and some want things to stay as it was back in the day so this whole feature is toggled by a new config value I added to a new config file called QualityOfLifeFeatures.cfg and the default value for this is false so no one needs to see any changes unless they want to.

If the feature is enabled then three new player commands become available to the players:

1. [toggleautoloot - This toggles between enabling and disabling the auto looting of gold for the player.  If this is enabled and the player kills a non-player mob that has gold on its corpse then the gold will automatically be placed in the players backpack rather than in the corpse's container.  If more than one player is involved in the kill then all involved players must have the auto loot feature enabled for them or the gold will simply spawn on the corpse.  If all players involved do have the feature enabled then the gold is even split among the players involved (and possibly party members - see below) with any remainder being lost to the ultima online gods.
2. [toggleautolootmessages - This toggles between enabling and disabling the player from receiving a message when auto looted gold is placed in their back pack.  The two messages possible are: "Your share of the gold from killing [NAME OF MOB] is [AMOUNT OF GOLD] gold." and "You loot the corpse of [NAME OF MOB] and place [AMOUNT OF GOLD] gold in your backpack." depending on if they were the only player involved in the kill or not.
3. [toggleautolootsharewithparty - This toggles between enabling and disabling sharing your auto looted gold with the party members of all players involved with the kill.  As stated above this will not share with party members unless all those involved in the kill have this enabled.  This will share with all players in all parties then.  Please note that not all party members need to have this enabled just all those involved in the kill of the mob that drops the gold.

Currently all of these new player settings default to a true value if the feature is enabled at the server level.

### DIFF
--- a/Config/QualityOfLifeFeatures.cfg
+++ b/Config/QualityOfLifeFeatures.cfg
@@ -1,0 +1,2 @@
+# Enable auto looting of gold
+AllowGoldAutoLoot=true
--- a/Config/QualityOfLifeFeatures.cfg
+++ b/Config/QualityOfLifeFeatures.cfg
@@ -1,2 +1,2 @@
 # Enable auto looting of gold
-AllowGoldAutoLoot=true
+AllowGoldAutoLoot=false
--- a/Scripts/Commands/AutoLoot.cs
+++ b/Scripts/Commands/AutoLoot.cs
@@ -1,0 +1,81 @@
+﻿using Server.Mobiles;
+
+namespace Server.Commands
+{
+	public class AutoLoot
+	{
+		[CallPriority(100)]
+		public static void Initialize()
+		{
+			if (Config.Get("QualityOfLifeFeatures.AllowGoldAutoLoot", false))
+			{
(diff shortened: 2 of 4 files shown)


noodleses:
I am not sure why PlayerMobile.cs says there are over 12k line changes.  These do not show up in VS for me.  If anyone has a suggestion let me know.

kamronbatman:
> I am not sure why PlayerMobile.cs says there are over 12k line changes. These do not show up in VS for me. If anyone has a suggestion let me know.

Try it like this:
https://github.com/ServUO/ServUO/pull/4982/files?diff=split&w=1

noodleses:
Thanks!  Yes that shows it much better

Voxpire:
I really like these features, however they do fall under the custom category; 
They would be perfect as a resource release on servuo.com

PR #4983, opened 2022-01-03
Xml tests

ruaduck:
XML Spawner Working for ServUO Master


Voxpire:
Will also be using this PR for reference soon :P

PR #4987, opened 2022-01-16
Fixing check around MobileIncoming calls

noodleses:
Fixing the issue with mobs disappearing after about an hour of ideal time.


Voxpire:
Thanks for this, it didn't fix the issue unfortunately. 
I did however manage to track it down and fix it!

Issue #4991, opened 2022-02-11
Repair Bench


KrystalGold:
The repair bench doesn't remove charges.

pyro17:
@Voxpire this is already merged, and it can be closed

PR #4992, opened 2022-02-13
RepairBenchFix

TrueUO:
Changes this

`if (m_Addon == null || !m_Addon.Deleted)
                return;`

To this

`if (m_Addon == null || m_Addon.Deleted)
                return;`

It should be (if the addon is null or deleted, return)

not (if the addon is null or NOT deleted, return) 

Takes care of: https://github.com/ServUO/ServUO/issues/4991


Voxpire:
Very nice, beat me to it! :D

PR #4993, opened 2022-02-16
Fix Mobile creation

kaczy93:
Mobile can already be deleted before delayed call will generate spawning loot, resulting in "inaccessible items".

### DIFF
--- a/Server/Item.cs
+++ b/Server/Item.cs
@@ -5858,7 +5858,13 @@ namespace Server
 				m_TypeRef = World.m_ItemTypes.Count - 1;
 			}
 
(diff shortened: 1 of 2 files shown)


Voxpire (changes requested):
This is a good catch! We should also check if the mobile is deleted after invoking the event.
Item seems to suffer the same issue.

kamronbatman (commented):
While this prevents the issue, it also breaks the assumption that the event will always fire when a mob is created. Is that ok? @Voxpire, what do you think?

kaczy93:
Is there anything else I can add to this issue or is it fine to merge it?

PR #5009, opened 2022-05-14
Listener and SocketOptions Fix.

UOShadowAge:
+ By default, IP Endpoint should only be the IP that the end user has configured themselves, otherwise All.
+ Added examples to the server config to make the options more transparent.


Voxpire:
These issues have been handled in p58-wip with the introduction of a more extensive server configuration.

Issue #4990, opened 2022-01-25
Ranger Mobile Script Error


GoldDraco72:
I noticed that there is a issue with Ranger, and potentially other mobiles, when you add a ranger to the world, it throws exceptions, no crash as there handled but the code is loading the layer twice!

            base.InitOutfit(); 
            AddItem(new Items.Shirt(Utility.RandomNeutralHue()));
            AddItem(new Items.LongPants(Utility.RandomNeutralHue()));
            AddItem(new Items.Bow());
            AddItem(new Items.ThighBoots(Utility.RandomNeutralHue()));

The Mob is first dressed via base.InitOutfit();

but then gets items added which are throwing the errors and end up not being equipped!

![image](https://user-images.githubusercontent.com/58917975/150917506-8e861a6a-1fe1-4f21-893f-4f749638cee3.png)

As you can see, the mobile is not being equipped with the thighboots, as coded, it is erroring out and reverting to the shoes added in the base.Init

PR #5012, opened 2022-06-12
Ranger Equipment fix, Additem delete

pyro17:
- Add parameter with default value to AddItem
- On Conflict, still log conflict, but delete the conflicting Item by default

- Delete the base boots of Rangers befor adding the new boots

Open Issue  #4990

### DIFF
--- a/Scripts/Mobiles/Bosses/Barracoon.cs
+++ b/Scripts/Mobiles/Bosses/Barracoon.cs
@@ -42,11 +42,11 @@ namespace Server.Mobiles
             Fame = 22500;
             Karma = -22500;
 
-            AddItem(new FancyShirt(Utility.RandomGreenHue()));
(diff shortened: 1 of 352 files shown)


Voxpire (changes requested):
This issue can be resolved by replacing any creature's AddItem usages with SetWearable.
SetWearable was introduced to resolve these issues, but it seems like not all creatures were updated to use it.
https://github.com/ServUO/ServUO/blob/master/Scripts/Mobiles/Normal/BaseCreature.cs#L5298

pyro17:
Didnt see / know of that method.

But looking at the method, it doesnt replace the item but the one that you want it to wear is placed inside the backpack.
Is that how it was intended? In my opinion the one that was "overwritten" should be either moved or deleted?

pyro17:
Updated the PR

Issue #5010, opened 2022-05-15
TeleporterRegions.xml missing exit from Despise Trammel


Lexshmyr:
Trammel Despise allows entry to the dungeon, but not allowing exit by walking. 

The file ./Data/TeleporterRegions.xml is missing content for leaving the dungeon through the main entrance/exit. There are builds where TeleporterRegions.xml was included (pub 57/57.1 and 58) while the former method  

I modified the file to add the following entry ahead of entrance code.
(New entry to allow exit from Despise to Trammel:)
>         <Teleporter>
>                <tiles from="(5588, 632, 30)" frommap="Trammel" to="(1296, 1082, 0)" tomap="Trammel" />
>                <tiles from="(5588, 631, 30)" frommap="Trammel" to="(1296, 1081, 0)" tomap="Trammel" />
>                <tiles from="(5588, 630, 30)" frommap="Trammel" to="(1296, 1080, 0)" tomap="Trammel" />
>        </Teleporter>
This is the Despise Entrance code and adding the above entry makes sense in proximity to this part:
>        <Teleporter>
>                <tiles from="(1296, 1082, 0)" frommap="Trammel" to="(5588, 632, 30)" tomap="Trammel" />
>                <tiles from="(1296, 1081, 0)" frommap="Trammel" to="(5588, 631, 30)" tomap="Trammel" />
>                <tiles from="(1296, 1080, 0)" frommap="Trammel" to="(5588, 630, 30)" tomap="Trammel" />
>        </Teleporter>
[TeleporterRegions.txt](https://github.com/ServUO/ServUO/files/8694251/TeleporterRegions.txt)

The attached file has the extension ".txt" because it won't allow dropping ".xml" file as an extension, so if you decide to just grab that file, you will have to rename it's extension.
The TeleporterRegions.xml file resides in the base ServUO directory under /Data

TrueUO:
https://github.com/ServUO/ServUO/blob/master/Scripts/Items/Internal/DespiseTeleporter.cs

The problem with this, and why it may be missing in the first place, is the exit teleporters need to be DespiseTeleporters to prevent people from leaving the dungeon with their Despise creatures.

PR #5013, opened 2022-06-12
Adding exit to trammel despise

pyro17:
- added exit teleporter tiles to the xml file

PR for Issue #5010

### DIFF
--- a/Data/TeleporterRegions.xml
+++ b/Data/TeleporterRegions.xml
@@ -797,6 +797,11 @@
 		<tiles from="(5573, 630, 42)" frommap="Trammel" to="(5503, 571, 51)" tomap="Trammel" />
 	</Teleporter>
 	<Teleporter>
+		<tiles from="(5588, 632, 30)" frommap="Trammel" to="(1296, 1082, 0)" tomap="Trammel" />
+		<tiles from="(5588, 631, 30)" frommap="Trammel" to="(1296, 1081, 0)" tomap="Trammel" />
+		<tiles from="(5588, 630, 30)" frommap="Trammel" to="(1296, 1080, 0)" tomap="Trammel" />
+	</Teleporter>
+	<Teleporter>
 		<tiles from="(1296, 1082, 0)" frommap="Trammel" to="(5588, 632, 30)" tomap="Trammel" />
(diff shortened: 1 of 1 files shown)


TrueUO:
https://github.com/ServUO/ServUO/blob/master/Scripts/Items/Internal/DespiseTeleporter.cs

The problem with this, and why it may be missing in the first place, is the exit teleporters need to be DespiseTeleporters to prevent people from leaving the dungeon with their Despise creatures.

pyro17:
Hmm maybe, but at the same time wouldnt it make more sense to only have 1 Teleport?

the other way would be to use the Evensink on the creatures.


			EventSink.OnEnterRegion += (e) => { if (!(e.NewRegion is DespiseRegion || Map != Map.Trammel)) MoveToWorld(new Point3D(5588, 632, 30), Map.Trammel); };

should be enough to keep them in (as example the entrance location) 

would need to unsubscribe on death / delete to clean it up.

That way you would have a way to keep them inside and use 1 teleport system.
And a default way to leave the dungeon

DocKnight86:
Oh I think that makes total sense. Just pointing out the issue, as it stands, with using normal teleporters at the exit.

TrueUO:
I think that makes a lot of sense.

I'm just pointing out the issue, as it stands, with using normal teleporters at the exit

pyro17:
Was just throwing that in as an alternative.

Question is, what would be the wanted / wished for way

Lexshmyr:
As it stands, the referenced code has no exit teleporters because the command to add the teleporters references a file that had it's name changed. That seemed to indicate it wasn't working - rather than some of the other pub 57.x->58 deco changes where directories were reorganized. 
The file from \Data\TeleportersOld.csv has the data to create those exit teleporters for Despise:
`# Teleporter X,Y,Z,Map,Dest X,Y,Z,Map,Back?
# Old Despise
5409,860,52,Trammel,5387,757,-3,Trammel,False
5409,858,52,Trammel,5387,755,-3,Trammel,False
5409,859,52,Trammel,5387,756,-3,Trammel,False`
This file was referenced in /Scripts/Commands/GenTeleporter.cs with the name "teleporters.csv".

I was not around for that change to deactivate the logic by changing the filename, but can only imagine there was some other issue that was not working related to all of this. It doesn't seem right that simply renaming the file back is the solution.

PR #5026, opened 2022-10-27
serial-optimization

Kimuji:
go to next serial value after a free value is found
that way you do one less World.FindItem/World.FindMobile 
each time because the value you found in the previous call will probably be used

### DIFF
--- a/Server/Serial.cs
+++ b/Server/Serial.cs
@@ -30,6 +30,7 @@ namespace Server
 						--m_NextMobileQueued._Value;
 
 					s = m_NextMobileQueued;
+					--m_NextMobileQueued._Value;
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


Kimuji:
Could be changed to a do{...}while(); too if you prefer

Kimuji:
I don't get the issue with gaps since you'll have gaps
when deleting mobiles and items.
You're the master, i'll keep this in my repo ;)

Voxpire:
```
for (var i = 0; i < 100; i++)
{
    var item = new Item();
    
    item.Delete();
}
```
In this case, all 100 item instances will have the exact same serial value, because every sequential construction will test the world dictionary for existence without incrementing the serial.
Your update would cause this code to create 100 new serials instead, leaving a gap of 100 unusable serials until the next server restart when the serials are reconstituted.

The dupe command will also create gaps if it fails to initialize the item properly.

PR #5027, opened 2022-11-07
Properties gump enum flag support

pyro17:
Properties gump enum flag support.

Works on enums with the Flags Attribute


pyro17:
Hmm will retry so it doesnt include all the old commits ;D

PR #5039, opened 2023-01-16
Fixing Reptalon|ColdDrake|CuSidhe Magery Mastery

LordZiru:
You can't train Reptalon|ColdDrake|CuSidhe to the "Magical Abilities" Magery Mastery and you must be able to. Just add MageryMastery | to line 97


LordZiru:
It is a simple fix.
just adding Magery  Mastery to Cold Drakes will add it to Cu sidhes and Reptalons too.

PR #5042, opened 2023-02-13
ScrollBinderExploitFix

TrueUO:
- If you add two 1.0 scrolls of transcendence to a scroll binder it will correctly delete the first one but when you target the other one it will create a new 2.0 SOT while leaving the last 1.0 SOT. This can allow users to "dupe" these scrolls. It only happens when you hit 2.0 exactly in a scroll binding deed.
- This small edit resolves the exploit.


TrueUO (commented):
-resolved

PR #5043, opened 2023-02-22
Remove the forced skillcap from the reward system

bittiez:
I don't see any reason the reward system should be forcing the skill cap. Other scripts that change the players skill caps dynamically are being overridden by this.


Iomega0318:
You know it actually took me a while to find this was the issue as my skill cap kept resetting when a player would log in after changing it in game, this was the cause.

PR #5057, opened 2023-05-11
add damage types to the book of masteries when discordance mastery is…

shroeder:
… enabled

On tick of despair, use selected damage type as the source of damage

image

from osi:
image


shroeder:
image

PR #5059, opened 2023-05-20
Critical Bug: Container.cs could access to bank outside of the bank b…

tr1age:
…ounds

This was triggered by client side that the server didn't check for. Issue here: https://discord.com/channels/@me/1079598263109427301/1109525125692407949


Voxpire:
The proposed fix does not fully address the issue, the storage logic must be changed in the core to be entirely effective.

PR #5072, opened 2023-11-10
fix: stepping onto the Paroxysmus Teleporter without a rope doesn't warn the player.

lilezek:
Looks like the else is just a mistake. Right now it only warns the player if the player doesn't have a backpack.


TrueUO:
Good catch. This exists on the main branches too. The whole OnMoveOver override function could be refactored a bit. That else and error message should probably be inside of the {} after the if (rope != null && !rope.Deleted)

TrueUO:
https://github.com/ServUO/ServUO/pull/5074

This will only update main. Not 57. So keep your PR. This is just how I think it should probably be done.

lilezek:
I'm closing it then.

PR #5074, opened 2023-11-11
ParoxTeleporterFix

TrueUO:
https://github.com/ServUO/ServUO/pull/5072

Based on PR 5072 for publish 57.

### DIFF
--- a/Scripts/Items/Internal/ParoxysmusTeleporters.cs
+++ b/Scripts/Items/Internal/ParoxysmusTeleporters.cs
@@ -40,9 +40,9 @@ namespace Server.Items
 
                     return base.OnMoveOver(m);
                 }
-            }
-            else
+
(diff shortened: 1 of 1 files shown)


TrueUO:
https://github.com/ServUO/ServUO/pull/5072

lilezek:
With this change the player won't get the message if they don't have a backpack. Is this intended? I'm not sure how can, a player,  lose their backpack at all.

TrueUO:
No, its just closer to the actual check for the backpack or not. Inside the bracket. 

Everything still falls through. Test it out.

TrueUO:
And you are exactly right. A players backpack should probably never be null, so nesting that message outside of the backpack check just didn't make sense. However, since it still falls through, unless it finds rope, you do still see the message. Just semantics. This also makes the changes to both the trammel and felucca teleporter.

PR #5080, opened 2024-04-27
Remove System.Drawing.Imaging references from ArtData

kaczy93:
Removed reference to System.Drawing.Imaging from ArtData.
Simplified it to only read dimensions and bounds (no access to pixel colors)

### DIFF
--- a/Server/Item.cs
+++ b/Server/Item.cs
@@ -1221,7 +1221,7 @@ namespace Server
 			}
 		}
 
-		public static Bitmap GetBitmap(int itemID)
+		public static ArtInfo GetBitmap(int itemID)
 		{
 			try
 			{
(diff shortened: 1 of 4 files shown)


aryman28:
Fcuk that xD


sob., 27 kwi 2024, 20:30 użytkownik kaczy93 ***@***.***>
napisał:

Server/Item.cs
```diff
@@ -1235,19 +1235,27 @@ public static Bitmap GetBitmap(int itemID)
...
+			{
+				xMin = bmp.Bounds.Left;
+				yMin = bmp.Bounds.Top;
+				xMax = bmp.Bounds.Right;
+				yMax = bmp.Bounds.Bottom;
+			}
 		}
 
-		public static Rectangle MeasureBound(Bitmap bmp)
+		public static Rectangle MeasureBound(ArtInfo bmp)
```
AsYlum-, line 1251: You may want to rename argument to something else as it's no longer bitmap `bmp`

Server/Item.cs
```diff
@@ -1235,19 +1235,27 @@ public static Bitmap GetBitmap(int itemID)
...
 		{
 			Measure(bmp, out var xMin, out var yMin, out var xMax, out var yMax);
 
 			return new Rectangle(xMin, yMin, xMax - xMin, yMax - yMin);
 		}
 
-		public static Size MeasureSize(Bitmap bmp)
+		public static Size MeasureSize(ArtInfo bmp)
```
AsYlum-, line 1258: You may want to rename argument to something else as it's no longer bitmap `bmp`

Server/Item.cs
```diff
@@ -1221,7 +1221,7 @@ public void SetVisibility(Mobile m, bool? state)
 			}
 		}
 
-		public static Bitmap GetBitmap(int itemID)
+		public static ArtInfo GetBitmap(int itemID)
```
AsYlum-, line 1224: Is it really GetBitmap() at this point? :)

PR #5081, opened 2024-04-29
Fix: Update Nails.cs Constructable to show correct sprite

Bohicatv:


kamronbatman:
Note to @Voxpire to copy this to p58-wip

PR #5098, opened 2024-11-28
Fixed BaseBoat Serialize 

kaczy93:
TillerMan can be null


Voxpire (changes requested):
This will actually break serialization because it will not write an expected value.
Writing a null entity is safe.
It would be safer to do:
```
writer.Write(TillerMan as IEntity);

TillerMan = reader.ReadEntity();
```

PR #5119, opened 2026-03-31
Bug/fixed pet justice

DaggerfordTechnologies:
Bug: In [PlayerMobile.cs:3973-3985]

../ServUO/Scripts/Mobiles/PlayerMobile.cs#L3973-L3985), the justice award block correctly resolves killer as the pet's owner (line 3900–3903), but then passed m (the raw FindMostRecentDamager result) to VirtueHelper.Award and all the follow-up message/effect calls. When a pet killed the murderer, m was the pet (BaseCreature), so the virtue went to the pet instead of the owner.

Fix: Replaced all m references inside the justice award block with killer, so the virtue, messages, and effects all target the owner. The killer.m_NextJustAward cooldown was already correct.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -4,7 +4,18 @@
       "Bash(./ServUO.exe -debug)",
       "Bash(dotnet build:*)",
       "Bash(dotnet --version)",
-      "Bash(echo $?)"
+      "Bash(echo $?)",
+      "Read(//c/code/Scripts/**)",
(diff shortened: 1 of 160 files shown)


Voxpire:
Please fork and then submit the fix separate from the rest of your customizations :)

DaggerfordTechnologies:
Sorry. My first time using github with other people. My mistake and chose the wrong repo.

